PR #62, opened 2014-09-20
Fix installed version detection in kvm.sh

graemechristie:
Fix the path for tests of installed versions by adding "/bin" to the package path. 

This resolves the issue where a KRE install that aborted before unpacking the Nuget package (because for e.g. the user ctrl-c'd during the package download) would be detected as installed and if attempt was made to install that version again, it would then fail with an error.


glennc:
This is just way to old to merge in now.

PR #104, opened 2014-11-21
Prevent duplicate active versions in kvm list

victorhurdugaci:
Fixes https://github.com/aspnet/kvm/issues/86

@graemechristie please review


graemechristie:
Looks good to me.

:shipit:

PR #115, opened 2014-12-16
Added the installation scripts from aspnet/Home

anurse:
Also deleted the kvmsetup.cmd because it is confusing and unnecessary (just run 'kvm.cmd setup' :))

Related to aspnet/homebrew-k#8


Praburaj:
@anurse - can we close this PR as we are not taking this forward?

PR #125, opened 2014-12-31
Fixing regression in curl command

Praburaj:
While making the below fix I accidentally removed the '-' parameter value being passed to the '-D' switch of curl command.
https://github.com/aspnet/kvm/commit/0656aeeab37383be661e2a6e0ddbf3252e409f86

'-D' indicates that the headers be written to a file. Adding the '-' back.

@Eilon


Eilon:
:shipit:

Issue #81, opened 2014-10-17
kvm upgrade -g fails with an error


muratg:
Repros with current kvm.ps1s from both Kvm/dev and Home/dev.

```
C:\>kvm upgrade -g
Determining latest version
There are no packages for platform '', architecture '' in the feed 'https://www.myget.org/F/aspnetvnext/api/v2'
Type 'kvm help' for help on how to use kvm.
```
kvm install latest -g works fine:
```
C:\>kvm install latest -g
Determining latest version
KRE-CLR-x86.1.0.0-beta1-10579 already installed.
Adding C:\Program Files\KRE\packages\KRE-CLR-x86.1.0.0-beta1-10579\bin to process PATH
```

Praburaj:
Fixed with https://github.com/aspnet/kvm/pull/127

PR #127, opened 2015-01-02
Fixing kvm-Global-upgrade

Praburaj:
When kvm upgrade is executed with a -g flag, today the code tries to fetch the latest version details from the REST url passing in the selectedRuntime and selectedArch explicitly.

These parameters have null values when this command is executed. Right thing is to have code exactly similar to Kvm-Upgrade but pass on global flag to Kvm-Install. This also avoid duplicate code on (Needs-Elevation) case. I verified this fixes the issue.

Fixes bug: https://github.com/aspnet/kvm/issues/81

@Eilon @muratg

### DIFF
--- a/src/kvm.ps1
+++ b/src/kvm.ps1
@@ -149,14 +149,7 @@ function Kvm-Global-Setup {
 function Kvm-Global-Upgrade {
   $Persistent = $true
   $Alias="default"
-  $versionOrAlias = Kvm-Find-Latest $selectedRuntime $selectedArch
-  If (Needs-Elevation) {
-    $arguments = "-ExecutionPolicy unrestricted & '$scriptPath' install '$versionOrAlias' -global $(Requested-Switches) -wait"
-    Start-Process "$psHome\powershell.exe" -Verb runAs -ArgumentList $arguments -Wait
-    Kvm-Set-Global-Process-Path $versionOrAlias
-    break
-  }
-  Kvm-Install $versionOrAlias $true
+  Kvm-Install "latest" $true
 }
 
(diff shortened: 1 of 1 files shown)


Eilon:
I totally don't know this code, sorry. And I can't read PowerShell. Not very sorry about *that* :smile:

Praburaj:
sure will wait for subject matter experts :)

graemechristie:
@Praburaj that change will only work if you already have elevated permissions. The bulk of the code in there is to handle the situation where the command is executed from a non-elevated prompt (and so invoke an elevated powershell instance). I'll have a look at #81 and commit an alternative fix.I don't suggest merging this PR.

graemechristie:
You can also remove the KVM-Global_Upgrade function and change the call in the switch statement to 
 `"upgrade 0"         {Kvm-Upgrade}`

(i.e. The Kvm-Upgrade and Kvm-Global-Upgrade functions are now the same, you don't need both)

Praburaj:
Thanks @graemechristie. There is still a difference between `KVM-Global-Upgrade` and `KVM-Upgrade` functions. Internally they call the `KVM-Install` with different flags. I can alternatively make the `KVM-Upgrade` to take in a bool flag indicating global upgrade or not and unify them.

Praburaj:
@graemechristie - updated the PR. Removed the method `Kvm-Global-Upgrade`.

suhasj:
reviewed thread and code :shipit:

Praburaj:
Thanks @graemechristie and @suhasj

PR #131, opened 2015-01-08
Fixes #130 by adding a hash check for the test nupkg

### LINKED ISSUE #130: Tests can fail if the downloaded kre package is corrupt
We download a KRE package to use during testing (to test installing from a local NuPkg) and if that file is corrupt, it causes the tests to fail in weird ways. We should do the following

1. If the file is present, check it against a known sha-256 hash. If the file does not match, delete it and redownload it.
1. If the file is not present, download it and check it against the known sha-256 hash.

I'll have a fix for this very shortly, just wanted to log it.

anurse:
@Praburaj and I were seeing some occasional corruption in the nupkg downloaded by the tests for testing `kvm install <nupkgPath>`. It is compounded by the fact that if the file exists with the expected name, the tests never redownload them. So I added a SHA-256 hash check in that ensures the package is valid. It's still possible for the tests to fail due to corruption, but now it should fail early and have a descriptive message. Also, re-running the tests will re-attempt the download.

Fixes #130

### DIFF
--- a/test/ps1/_Common.ps1
+++ b/test/ps1/_Common.ps1
@@ -52,3 +52,25 @@ function GetKreName {
     param($clr, $arch, $ver = $TestKreVersion)
     "KRE-$clr-$arch.$ver"
 }
+
+# Borrowed from kvm itself, but we can't really use that one so unfortunately we have to use copy-pasta :)
+# Modified slightly to take in a Proxy value so that it can be defined separately from the Proxy parameter
+function Add-Proxy-If-Specified {
+param(
+  [System.Net.WebClient] $wc,
+  [string] $Proxy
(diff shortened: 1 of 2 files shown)


Praburaj:
Talked with @anurse - we are going to use `WebClient` like `kvm.ps1` to download the `nupkg` file since we have not seen corruption issues while using `kvm`. Since we are not very sure what's causing the corruption we will monitor it for some more time for flakiness.

anurse:
Added the WebClient code now. We should be using the same code for downloading in both kvm and the tests.

Praburaj:
Looks good :shipit: . You will have to rebase your changes.

PR #132, opened 2015-01-08
Updating kvm install command line help to take proxy.

Praburaj:
Also updated the help text with an example.

This will also address the concern raised in https://github.com/aspnet/Home/issues/72.

/CC @anurse for review.


anurse:
:shipit:

PR #134, opened 2015-01-13
Add kvm.sh tests

anurse:
As before, once this is up and running, there will be a separate PR to turn on pushing of KVM over to aspnet/Home

I've set up a new KVM build on the TeamCity server too, it now builds in a separate project where we can have configurations for Windows, Mac, and Linux in order to build on all platforms.


NuGet.Config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/" />
+    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
```
anurse: I'll clean this up before merge.

src/kvm.sh
```diff
@@ -308,7 +310,7 @@ kvm()
                 printf "$format" "Alias" "Name"
                 printf "$format" "-----" "----"
                 for _kvm_file in $(find "$KRE_USER_HOME/alias" -name *.alias); do
-                    local alias="$(basename $_kvm_file | sed 's/.alias//')"
+                    local alias="$(basename $_kvm_file | sed 's/\.alias//')"
```
anurse: I was seeing an issue with `kvm alias` that would list the alias names as `default.alias` instead of just `default`. This fixes that (and the tests now verify it!)

src/kvm.sh
```diff
@@ -365,10 +367,12 @@ kvm()
             # Z shell array-index starts at one.
             local i=1
             local format="%-20s %s\n"
-            for _kvm_file in $(find "$KRE_USER_HOME/alias" -name *.alias); do
-                arr[$i]="$(basename $_kvm_file | sed 's/.alias//')/$(cat $_kvm_file)"
-                let i+=1
-            done
+            if [ -d "$KRE_USER_HOME/alias" ]; then
```
anurse: If the alias directory was missing, we'd get weird errors. There's no reason not to allow it to be missing though, that just implies no aliases are defined.

test/sh/run-tests.sh
```diff
@@ -0,0 +1,136 @@
...
+popd > /dev/null
+
+# Determine root
+pushd $SCRIPT_DIR/../.. > /dev/null
+REPO_ROOT=$(pwd)
+popd > /dev/null
+
+# Default variable values
+[ -z "$TEST_WORK_DIR" ]     && export TEST_WORK_DIR="$(pwd)/testwork"
+[ -z "$TEST_SHELLS" ]       && export TEST_SHELLS="bash zsh"
```
anurse: Any other shells we should test in? It looks like kvm doesn't really work in plain "sh" (unless your distro has aliased "sh" to "bash").

test/sh/urchin.sh
```diff
@@ -0,0 +1,267 @@
+#!/bin/bash
```
anurse: This is a modified copy of tlevine/urchin to add some TeamCity-related things. I can look at contributing it back. It's licensed under the BSD license, so we should be OK to use it (https://github.com/tlevine/urchin/blob/f01869fb97687794cc415bc7a5357b03a2afc467/LICENCE)

@Eilon - You see any issues with us doing this? It's a nice, small, shell script so it's not too scary to have our own copy for the near-term.

anurse:
/cc @Praburaj

Not sure who, if anyone, is familiar with kvm.sh specifically. Feel free to loop others in.

src/kvm.sh
```diff
@@ -81,7 +81,8 @@ _kvm_download() {
         return 1
     fi
 
-    mkdir -p "$kreFolder" > /dev/null 2>&1
+    [ -e "$KRE_USER_PACKAGES" ] || mkdir "$KRE_USER_PACKAGES" > /dev/null 2>&1
```
Praburaj: Can we move the directory creation commands after the http command below. We right now end up creating the package folder even if the package is not found in the feed. I remember there is an issue on that.

src/kvm.sh
Praburaj: :+1:

src/kvm.sh
```diff
@@ -365,10 +367,12 @@ kvm()
             # Z shell array-index starts at one.
             local i=1
             local format="%-20s %s\n"
-            for _kvm_file in $(find "$KRE_USER_HOME/alias" -name *.alias); do
-                arr[$i]="$(basename $_kvm_file | sed 's/.alias//')/$(cat $_kvm_file)"
-                let i+=1
-            done
+            if [ -d "$KRE_USER_HOME/alias" ]; then
+                for _kvm_file in $(find "$KRE_USER_HOME/alias" -name *.alias); do
+                    arr[$i]="$(basename $_kvm_file | sed 's/.alias//')/$(cat $_kvm_file)"
```
Praburaj: Does this also need the `sed` pattern escape for the `.alias`?

test/sh/run-tests.sh
```diff
@@ -0,0 +1,136 @@
...
+requires awk
+
+USAGE="usage: $0 [<options>] <test directory>"
+
+runtests_help() {
+    cat <<EOF
+
+$USAGE
+
+-t          Writes TeamCity status messages to the output.
```
Praburaj: We should probably add some notes to `readme.md` on how to run the tests in general for both windows and non-windows.

test/sh/tests/install/Installing from nupkg unpacks nupkg as expected.sh
```diff
@@ -0,0 +1,13 @@
...
+source $KVM
+
+# Install the nupkg
+kvm install $KRE_NUPKG_FILE
+
+[ -d "$KRE_USER_HOME/packages/$KRE_NUPKG_NAME" ] || die "unable to find installed KRE"
+
+pushd "$KRE_USER_HOME/packages/$KRE_NUPKG_NAME" 2>/dev/null 1>/dev/null
+[ -f bin/k ] || die "KRE did not include 'k' command!"
+[ -f bin/klr ] || die "KRE did not include 'klr' command!"
```
Praburaj: Do we need check for klr?

test/sh/common.sh
```diff
@@ -0,0 +1,52 @@
...
+has() {
+	type "$1" > /dev/null 2>&1
+    return $?
+}
+
+requires() {
+	if ! has $1; then
+		die "Missing required command: $1"
+	fi
+}
```
Praburaj: Is `file.nupkg` used anywhere? (I can't seem to write a command over `file.nupkg`).

test/sh/tests/install/Installing from nupkg unpacks nupkg as expected.sh
Praburaj: It does not hurt. This is good.

test/sh/run-tests.sh
davidfowl: fish shell

test/sh/common.sh
anurse: No, it was a bogus file. I deleted it.

test/sh/tests/install/Installing from nupkg unpacks nupkg as expected.sh
anurse: Yeah, these will need to be updated when the renames land anyway.

src/kvm.sh
anurse: Probably! I'll add that.

src/kvm.sh
anurse: Not easily. The curl command downloads to a path that is under this directory so it needs to exist before that command runs. If we do want to do that, we can investigate it but it should probably be a separate change from this one.

test/sh/chester
```diff
@@ -0,0 +1,319 @@
+#!/bin/bash
```
anurse: Instead of urchin, we're now using this quick-n-dirty test harness I whipped up. It's pretty basic.

test/sh/chester
```diff
@@ -0,0 +1,319 @@
...
+
+	FORMAT="${Pur}help ${RCol}:  ${Yel}%-20b${RCol} %b\n"
+	printf "$FORMAT" "-h" "Print this help message"
+	printf "$FORMAT" "-t" "Report status using TeamCity-style status messages"
+	printf "$FORMAT" "-v" "Report status using TeamCity-style status messages"
+	printf "$FORMAT" "-s <shell>" "Run the tests in ${Yel}<shell>${RCol}, defaults to '/bin/sh'"
+    printf "$FORMAT" "-n <name>" "When writing TeamCity status message, use <name> as the outer-most test suite name"
+}
+
+# Main entry point (after argument parsing)
```
anurse: This is where the fun begins if you want to trace through the code. Most of the code outside of this function call tree is argument parsing.

test/sh/run-tests.sh
anurse: fish isn't a POSIX-compliant shell. If we want to support it, we'll have to write a new KVM.fish for it.

anurse:
Ping. Going to merge this soon so I can test out the actual TeamCity build.

Praburaj:
Looks good :shipit:

PR #135, opened 2015-01-15
Fixing the typo in the command line option -persistant

Praburaj:
command line help seems to be good. However the parsed option value seems to be incorrect.

This was reported in bug # https://github.com/aspnet/kvm/issues/89

@anurse


anurse:
:shipit:

PR #141, opened 2015-01-20
Remove global install and get the ps1 build working again

anurse:
This re-enables the PS1 tests. A separate PR will be coming to re-enable the SH tests later this afternoon and that PR will also enable automatic pushing to aspnet/Home.

Also, we removed global install from 'dotnetsdk'. There will be a proper Windows MSI to do global installs and the dotnetsdk global install risks colliding with that and causing confusion (if you install something one way, can you remove it the other way? etc.)

### DIFF
--- a/src/dotnetsdk.ps1
+++ b/src/dotnetsdk.ps1
@@ -3,19 +3,14 @@ param(
   [string] $Command,
   [string] $Proxy,
   [switch] $Verbosity = $false,
-  [alias("g")][switch] $Global = $false,
   [alias("p")][switch] $Persistent = $false,
   [alias("f")][switch] $Force = $false,
   [alias("r")][string] $Runtime,
+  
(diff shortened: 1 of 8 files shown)


src/dotnetsdk.ps1
```diff
@@ -3,19 +3,14 @@ param(
...
   [alias("arch")][string] $Architecture,
   [switch] $X86 = $false,
-  [switch] $Amd64 = $false,
-  #deprecated
-  [switch] $X64 = $false,
-  #deprecated
-  [switch] $Svr50 = $false,
-  #deprecated
-  [switch] $Svrc50 = $false,
+  [alias("amd64")][switch] $X64 = $false,
```
anurse: These can be aliases now since we know the exact syntax expected in the final package ID format.

PR #145, opened 2015-01-21
restore dotnetsdk.sh tests and build

anurse:
Mostly updating the tests. Also, the change I made to correct alias display didn't get merged earlier (not sure why) so it's back here (that's the only actual change to `dotnetsdk.sh` itself)


anurse:
This will probably miss the beta3 snap, but I can port it over to the release branch when it's ready, it's a small change :)

test/sh/run-tests.sh
```diff
@@ -20,8 +20,8 @@ popd > /dev/null
 [ -z "$TEST_APPS_DIR" ]     && export TEST_APPS_DIR="$REPO_ROOT/test/apps"
 
 # This is a KRE to use for testing various commands. It doesn't matter what version it is
-[ -z "$DOTNET_TEST_VERSION"]   && export DOTNET_TEST_VERSION="1.0.0-beta1"
-[ -z "$DOTNET_NUPKG_HASH" ]    && export DOTNET_NUPKG_HASH="5fb3d472166f89898631f2a996f79de727f5815f"
+[ -z "$DOTNET_TEST_VERSION"]   && export DOTNET_TEST_VERSION="1.0.0-beta3-10935"
```
Praburaj: test will start failing when this package is deleted from the feed?

Praburaj:
Do we temporarily disable the test until we have a package in master feed that will be available for long? 

Apart from that :shipit:

anurse:
You're absolutely right about the tests failing when the package is deleted. However, I'd like to keep the test around if possible. Since it requires a small change to update the version, I think we can keep it as-is until we have a package in the master feed. Alternatively, I could look at putting a package up for purely testing purposes that wouldn't go away. I'll investigate some of those options when merging.

anurse:
Also, I'll actually be merging this in to release, since we've already snapped for beta3. @Eilon / @glennc  any concerns there?

anurse:
I've pinned the versions we're using so they won't be removed from aspnetvnext. Once there's a build on aspnetmaster, I'll unpin them and switch to that.

Eilon:
Approved for beta3.

PR #147, opened 2015-01-27
Rename back to kvm

anurse:
We aren't quite ready to finalize the name yet so we'll go back to kvm for now :)

/cc @muratg @Eilon @davidfowl @ChengTian


src/dotnetsdk.sh
```diff
@@ -1,416 +0,0 @@
-# dotnetsdk.sh
```
anurse: Guh. Another Rename+Edit turned into a Delete+Add :(

Eilon:
I don't really know how to review this, so as far as I'm concerned, if all the tests pass, then check it in :smile:

anurse:
Yeah... it's basically a big pile of renames...

muratg:
Looks good.   :shipit:

PR #152, opened 2015-02-04
Ngen Roslyn libraries on kvm install

shhsu:


src/k-ngen.ps1
```diff
@@ -0,0 +1,31 @@
...
+  Write-Error "Installation does not understand architecture $architecture, skipping ngen..."
+  Exit -1
+}
+
+$regHive = [Microsoft.Win32.RegistryHive]::LocalMachine
+$regKey = [Microsoft.Win32.RegistryKey]::OpenBaseKey($regHive, $regView)
+$frameworkPath = $regKey.OpenSubKey("SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full").GetValue("InstallPath")
+$ngenExe = Join-Path $frameworkPath 'ngen.exe'
+
+foreach ($bin in Get-ChildItem $runtimeBin -Filter "Microsoft.CodeAnalysis*.dll") {
```
davidfowl: I believe you can just ngen Microsoft.CodeAnalysis.CSharp and it will do the rest.

davidfowl:
Won't this prompt for admin?

src/k-ngen.ps1
shhsu: I see, would there be a negative impact to ngen all 4 of them?

shhsu:
Yes this would prompt for admin exactly when this line is executed
https://github.com/aspnet/dotnetsdk/pull/152/files#diff-2318c4cb7c76b821e5705ddcd40aa6b3R390
This is also the reason I have created k-ngen.ps1 because I seem to get 4 admin prompts if I execute the ngen command for the 4 files in 4 Start-Process calls

jhawk42:
@davidfowl are you ok with @shhsu making this change in kvm install? If so, lets move ahead with it.

davidfowl:
Nope, I think it needs more testing. How will this affect our builds scripts? Does it gracefully do nothing and warn if I cannot elevate? What if I'm automating install of the kvm (like every one of our builds)? Is there a way to not prompt? 

Did you try this change with our build scripts to see how it affects things?

anurse:
I can look at rebase this change on top of #156 once it's in and review it. I'm concerned about the Admin prompt. We should put this behind a switch and disable it by default. KVM is extremely intentional about never prompting for admin (since we removed global install), so I'd rather not add a new admin prompt...

anurse:
Did a quick sync-up with @davidfowl . We'll do this behind a switch (something like "kvm install -NGenRoslyn" but with a better name :)). KVM install is very often done as part of build scripts and we don't want to force everyone to go through and add a switch to disable this, so we'll make it opt-in.

anurse:
And a note on timing, #156 should go in today (doing final testing on Win7). I can handle merging and integrating this PR once that's done. So I expect we can have this in fairly soon.

shhsu:
I can't get the 3 way merge to work properly. At this point I would rather abandon this PR rather than trying to rebase it. I will submit a new PR soon

shhsu:
https://github.com/aspnet/dotnetsdk/pull/158

PR #156, opened 2015-02-06
Major overhaul to kvm.ps1

anurse:
I **HIGHLY** recommend just scanning the new version of kvm.ps1. The diff is a bit weird.

The tests all pass with only minor adjustments. I also added a few new tests.

Most of the changes are around restructuring the commands into logical units and making it easier to add new commands with help messages. Here's a brief summary:

* Commands are now added by just adding a new function that starts `kvm-[commandname]`. Help text is derived from PowerShell Doc Comments and Parameter attributes.

image

* Help text has been redesigned to be more consistent with other command-line tools:

image

image

* `kvm unalias` has been deprecated (but not removed) in favor of using `kvm alias -d`, which better aligns with other command line tools. `nvm` does have an unalias command though, so I'm open to keeping it around if people think it looks better.

image

* Architecture switches (`-amd64`, `-x64`, `-x86`) have been deprecated (but not removed) in favor of the, slightly more scriptable, `-arch [arch]` parameter

image

* `kvm list` learned a new switch `-PassThru` which returns raw, unformatted, PowerShell arguments for scripting
* `kvm` learned a new command `kvm name` which outputs the full name of a runtime given a version, architecture and runtime flavor

image

* Added a lot more debug spew, when the user requests it:

image

* Made `kvm-install` a little less confusing by removing duplication between the "install from nupkg path" and "install from feed" code paths.
* `kvm upgrade` can now upgrade any alias, not just `default` (though it defaults to `default`, that's the default alias it defaults to, by default). If the alias specifies a different runtime or architecture, that value will be preserved. For example, if I have `default-coreclr` pointing to my default coreclr build and I run `kvm upgrade default-coreclr`, it is updated to the latest coreclr build

/cc @davidfowl @Praburaj @suhasj 

kvm.sh is NOT affected at this time.


makefile.shade
```diff
@@ -35,4 +56,9 @@ default KVM_DEPLOY_BRANCH="${Environment.GetEnvironmentVariable("KVM_DEPLOY_BRAN
     exec program='/bin/bash' commandline="${Path.Combine(TEST_DIR, "sh", "run-tests.sh")} ${IsTeamCity?"-t ":""}-v" workingdir="${Path.Combine(TEST_DIR, "sh")}"
 
 #push-kvm target='deploy' if='DEPLOY_KVM == "1" && !String.IsNullOrEmpty(KVM_DEPLOY_REPO) && !String.IsNullOrEmpty(KVM_DEPLOY_BRANCH)'
-	push-kvm repo="${KVM_DEPLOY_REPO}" branch="${KVM_DEPLOY_BRANCH}" files="kvm.cmd;kvm.ps1;kvm.sh" baseMessage=":arrow_up: kvm.ps1, kvm.cmd, kvm.sh"
+	push-kvm repo="${KVM_DEPLOY_REPO}" branch="${KVM_DEPLOY_BRANCH}" files="${Path.Combine(TARGET_DIR, "kvm.cmd")};${Path.Combine(TARGET_DIR, "kvm.ps1")};${Path.Combine(TARGET_DIR, "kvm.sh")}" baseMessage=":arrow_up: kvm.ps1, kvm.cmd, kvm.sh"
+
+functions @{
+    string E(string key) { return Environment.GetEnvironmentVariable(key); }
+    void E(string key, string value) { Environment.SetEnvironmentVariable(key, value); }
```
Praburaj: Can we have names like `GetEnvironment`, `SetEnvironment` to these methods. Its not very obvious on what these are on a quick look.

src/kvm.ps1
```diff
@@ -1,188 +1,364 @@
...
+if($BuildVersion.StartsWith("{{")) {
+    # We're being run from source code rather than the "compiled" artifact
+    $BuildVersion = "HEAD"
+}
+$FullVersion="$ProductVersion-$BuildVersion"
+
+Set-Variable -Option Constant "CommandName" ([IO.Path]::GetFileNameWithoutExtension($ScriptPath))
+Set-Variable -Option Constant "CommandFriendlyName" "K Runtime Version Manager"
+Set-Variable -Option Constant "DefaultUserDirectoryName" ".k"
+Set-Variable -Option Constant "OldUserDirectoryName" ".kre"
```
Praburaj: We might want to file tracking bugs to remove these migration tips.

src/kvm.ps1
```diff
@@ -1,188 +1,364 @@
...
+    _WriteDebug "Saving output to '$__TeeTo' variable"
+    Set-Variable -Name $__TeeTo -Value "" -Scope Global -Force
 }
 
-# Get or calculate userHome
-$userHome = (getenv $UserHomeEnvVar)
-if(!$userHome) { $userHome = $env:USERPROFILE + "\$RuntimeFolderName" }
-$userRuntimesPath = $userHome + "\runtimes"
+# Commands that have been deprecated but do still work.
+$DeprecatedCommands = @("unalias")
```
Praburaj: Tracking bug to remove this switch as well.

src/kvm.ps1
```diff
@@ -199,592 +375,787 @@ param(
...
-    $script:exitCode = 1 # Return non-zero exit code for scripting
-  } else {
-    $aliasValue = (Get-Content ($userHome + "\alias\" + $name + ".txt"))
-    Console-Write "Alias '$name' is set to $aliasValue"
-  }
+<#
+.SYNOPSIS
+    Lists and manages aliases
+.PARAMETER Name
+    The name of the alias to read/write/delete
```
Praburaj: `write` => `create` ?

src/kvm.ps1
```diff
@@ -199,592 +375,787 @@ param(
...
-  if (Test-Path -literalPath "$aliasPath") {
-      Console-Write "Removing alias $name"
-      Remove-Item -literalPath $aliasPath
-  } else {
-      Console-Write "Cannot remove alias, '$name' is not a valid alias name"
-      $script:exitCode = 1 # Return non-zero exit code for scripting
-  }
+<#
+.SYNOPSIS
+    Upgrades an alias to the latest version of the runtime
```
Praburaj: I feel this synopsis may not be accurate. This actually installs the latest runtime from the feed + points the alias to this.

makefile.shade
anurse: Sure, I just copied these from k-standard-goals (since using all of k-standard-goals is overkill for this makefile)

src/kvm.ps1
anurse: Good point

src/kvm.ps1
anurse: :+1:

src/kvm.ps1
anurse: :+1:

src/kvm.ps1
```diff
@@ -199,592 +375,787 @@ param(
...
+.SYNOPSIS
+    Gets the full name of a runtime
+.PARAMETER VersionOrAlias
+    The version or alias of the runtime to place on the PATH
+.PARAMETER Architecture
+    The processor architecture of the runtime to place on the PATH (default: x86, or whatever the alias specifies in the case of use-ing an alias)
+.PARAMETER Runtime
+    The runtime flavor of the runtime to place on the PATH (default: clr, or whatever the alias specifies in the case of use-ing an alias)
+#>
+function kvm-name {
```
Praburaj: This is going to be useful while specifying `--runtime` switch for `kpm bundle`. I'm wondering instead of having this command - can the `kvm list` replace `Version` with `Name` as `version` information is just a subset of `Name`. We can avoid an additional command as well provide more useful information in `kvm list`.

Praburaj:
>kvm.sh is NOT affected at this time.

Can we just add the `{{AUTHORS}}` and `{{PRODUCTVERSION}}` parts to `kvm.sh` for now?

Praburaj:
PR looks pretty :smile:

src/kvm.ps1
anurse: I'd like a command that just outputs the name, because you can run something like `kvm name default | clip` and it will put the name on your clipboard for pasting later. However, I do like the idea of merging it with list. Perhaps allow an optional name after list: `kvm list default` and then a switch or something to only show the name: `kvm list -n default`.

I'll investigate that briefly and post back with my results.

anurse:
Snipping Tool + Copy/Paste on Github is wonderful :). You can paste an image straight out of your clipboard into a PR, no need to upload it anywhere or even save it to a file.

I can add those variables to kvm.sh.

Praburaj:
After adressing these comments it looks :shipit: to me.

anurse:
Cool, I'm doing some testing on Win7 because I think the auto-help features depend on a later version of powershell. I want it to at least gracefully fail in those cases :)

anurse:
Looks fine on Windows 7 with our minimum required PowerShell version of 4.0. Merging.

PR #158, opened 2015-02-10
Ngen Roslyn libraries on kvm install

shhsu:
Current behavior: Since ngen requires admin role. If the user is running in elevated mode, no prompt would occur. Otherwise, a prompt would come up and ask to run in elevated mode.

### DIFF
--- a/src/kvm.ps1
+++ b/src/kvm.ps1
@@ -551,6 +551,43 @@ SET "PATH=$newPath"
     }
 }
 
+function Ngen-Library(
+    [Parameter(Mandatory=$true)]
+    [string]$runtimeBin,
+
+    [ValidateSet("x86","x64")]
+    [Parameter(Mandatory=$true)]
+    [string]$architecture) {
+
+    if ($architecture -eq 'x64') {
+        $regView = [Microsoft.Win32.RegistryView]::Registry64
+    }
+    elseif ($architecture -eq 'x86') {
+        $regView = [Microsoft.Win32.RegistryView]::Registry32
+    }
+    else {
+        _WriteOut "Installation does not understand architecture $architecture, skipping ngen..."
+        return
+    }
+
+    $regHive = [Microsoft.Win32.RegistryHive]::LocalMachine
+    $regKey = [Microsoft.Win32.RegistryKey]::OpenBaseKey($regHive, $regView)
+    $frameworkPath = $regKey.OpenSubKey("SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full").GetValue("InstallPath")
+    $ngenExe = Join-Path $frameworkPath 'ngen.exe'
+
+    $ngenCmds = ""
+    foreach ($bin in Get-ChildItem $runtimeBin -Filter "Microsoft.CodeAnalysis.CSharp.dll") {
+        $ngenCmds += "$ngenExe install $($bin.FullName);"
+    }
+
+    $ngenProc = Start-Process "$psHome\powershell.exe" -Verb runAs -ArgumentList "-ExecutionPolicy unrestricted & $ngenCmds" -Wait -PassThru
+}
+
+function Is-Elevated() {
+    $user = [Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()
+    return $user.IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator")
+}
+
 ### Commands
 
 <#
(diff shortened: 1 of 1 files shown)


anurse:
Thanks for updating, I was going to do that but other things jumped up :). Can we do this without the "k-ngen.ps1" script? We really want kvm to be self-contained and have no dependencies. I was able to spike out something where I generated a temporary script file containing the `ngen` commands and then elevated powershell to run that script (in order to avoid multiple prompts).

shhsu:
Talked with @pranavkm
To integrate change with the build script we should pass in -nonative when installing CLR flavor of XRE

pranavkm:
We use the `-nonative` flag on the CI while installing the CoreCLR KRE packages - primarily to save time building. We can additionally pass the flag when installing CLR KRE once this change goes in.

anurse:
I was going to suggest using a `-Native` switch to enable NGenning on CLR. That way, existing scripts don't need to be modified (there are scripts in the general public as well, not just our build scripts).

The behavior would be:
* On CoreCLR, crossgen by default, use `-NoNative` to disable (`-Native` has no effect)
* On CLR, NO ngen by default, use `-Native` to enable (`-NoNative` has no effect)

It's a little odd, but I think it's the cleanest way to do it for now. I'm a little nervous about suddenly changing install to require admin with only an opt-out.

shhsu:
@anurse 
So this is the problem I was running into:

Admin prompt would appear each time the "Start-Pross / runAs" line is called for operations that needs elevated mode
I.E. Start-Process "$psHome\powershell.exe" -Verb runAs -ArgumentList "-ExecutionPolicy unrestricted & ***** " -Wait -PassThru

If we were to ngen multiple files I would have to call this line multiple lines to elevate multiple times and the user would be prompted multiple times.

The separate k-ngen.ps1 script is the only workaround I can think of at the time.

anurse:
You can generate a script that will do all the NGenning dynamically, and pass it to a single PowerShell invocation (using EncodedCommand: https://technet.microsoft.com/en-us/library/hh847736.aspx, or writing a temporary script file)

shhsu:
Sure, I will modify my script to get rid of k-ngen.ps1

On the -noNative side. This is basically the same deal as when we add in crossgen for CoreCLR. The installation process is changed unless the user opt out.

anurse:
Yeah, but `crossgen` doesn't require elevation, `ngen` does. That means that everyone who has scripted `kvm install` will suddenly have their scripts halt waiting for user input, which will be bad in build servers, etc.

I really think we should make this opt-in due to the admin requirement.

shhsu:
Would it help if I add a timeout plus error message as a temporary solution until all the build server has adapted to the changes?

I would think having to communicate to our future consumer of our script about this inconsistent opt-in opt-out logic is no less hassle than fixing exiting build system. And our interface definition is more of a permanent thing.

anurse:
I'm just really concerned about requiring admin to install a KRE under default conditions. We are explicitly designing this runtime to be installable without Admin privileges, but now I have to remember to add "-NoNative" to avoid elevation. I'd really much rather the native compilation be an opt-in process I can choose to use if I need the performance improvement. By default, we should only perform optimizations we can do without Admin privileges; other optimizations should be opt-in.

shhsu:
I see. If it is a concern and we would not want the user require admin privileges to install by default. Then we may need to require user to opt in on optimization for CLR flavor.

Just unsure about the priority of things. @davidfowl @jhawk42 , can you chime in?

davidfowl:
@anurse is right, this'll have to be opt in. We can do it automagically for the installers we have.

jhawk42:
@shhsu @anurse @davidfowl 
I think we want all groups of users (developers on their laptops/dev machines, folk installing into CI build machines, IT installing on web servers, cloud providers preinstalling K on their cloud servers) all  to fall into the pit of success of ease of installation and performance.

I agree that we don't want to needlessly prompt to elevate i.e. is one of the users above is just installing into their /users/foo directory or some random directory. That said we should looks for places to help these users fall into that pit of success and to ensure they have a great experience with the K product.

Here are some ideas for kvm install of K for DesktopCLR
* In powershell if kvm.ps1 is already running elevated (see link below to check in powershell) we should consider doing -native by default. Link: http://blogs.msdn.com/b/virtual_pc_guy/archive/2010/09/23/a-self-elevating-powershell-script.aspx
* Is there is a switch in kvm install to install globally i.e. program files, I assume we have to elevated already? if so, we should consider defaulting to -native
* When running non-elevated and installing K for Desktop we should consider provide user with console hint/information that to get "optimum performance" (someone like Hunter will have better words) they should consider adding -Native  and running kvm install elevated. Just a hint. Not a requirement.

anurse:
I'm certainly ok with adding the message. We removed global install using KVM in beta3 (in favor of using MSIs to do global install eventually), so there is currently no other reason for KVM to ever be run elevated.

I would probably be OK with doing NGen if we detect we are already admin. @davidfowl, you agree?

shhsu:
Updated as required

davidfowl:
Those suggestions sound great! It's non intrusive if elevated and warns that the user when it can't be done with a hint

shhsu:
Yes, that's what the code currently do now. Any feedback?

davidfowl:
@Praburaj can you give this a try?

Praburaj:
@davidfowl that seems to work.

shhsu:
@davidfowl can we check this in?

Praburaj:
Looks like :shipit: to me.

shhsu:
rebased

PR #162, opened 2015-03-03
how do I windows dir separator?

anurse:
In my defense, I... I... I...

I have no defense :(.

This is the cause of having multiple active KREs sometimes.

/cc @Praburaj


Praburaj:
:shipit:

PR #164, opened 2015-03-05
Update kvm.ps1

glennc:
Kudu fails with the restriction on 4. Tested the help commands on 3 and they appear to be fine.


davidfowl:
:shipit:

Issue #165, opened 2015-03-06
"kvm install [alias] -r coreclr" no longer works


DamianEdwards:
The ability to install an alternate runtime or architecture KRE of a given aliased KRE no longer works.

e.g.
`kvm install default -r coreclr` : installs the coreclr variant of the current default KRE

anurse:
Fixed in dev

PR #173, opened 2015-03-10
restore 'install <alias>' functionality and add better error handling

anurse:
We should now stop on the first error (apparently the outer try...catch does that? I barely know anything about PoSH error handling :)). Also, I trap a few well-known error conditions

Also restored `dnvm install <alias>` as it turned out to be a simple fix.

/cc @NTaylorMullen @glennc @davidfowl

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -100,6 +100,9 @@ $ExitCodes = @{
     "AliasDoesNotExist"         = 1001
     "UnknownCommand"            = 1002
     "InvalidArguments"          = 1003
+    "OtherError"                = 1004
+    "NoSuchPackage"             = 1005
+    "NoRuntimesOnFeed"          = 1006
 }
 
 $ColorScheme = $DnvmColors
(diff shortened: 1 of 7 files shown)


NTaylorMullen:
Looks good to me but I don't know nearly enough powershell :cry:

src/dnvm.ps1
```diff
@@ -449,7 +457,12 @@ function Download-Package(
     $wc = New-Object System.Net.WebClient
     Apply-Proxy $wc -Proxy:$Proxy
     _WriteDebug "Downloading $url ..."
-    $wc.DownloadFile($url, $DestinationFile)
+    try {
+        $wc.DownloadFile($url, $DestinationFile)
+    } catch {
+        $Script:ExitCode = $ExitCodes.NoSuchPackage
+        throw "Could not find $runtimeFullName.$Version on feed: $Feed"
```
davidfowl: Can this null ref? `runtimeFullName`

DamianEdwards:
#165

src/dnvm.ps1
anurse: It's not a property access :). It's rendering `{runtimeFullName}.{Version}`

PR #178, opened 2015-03-11
Crossgen output will not start a new process and pop out a new window

shhsu:
Response to https://github.com/aspnet/dnvm/issues/176

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -1054,13 +1054,18 @@ function dnvm-install {
         }
         elseif ($Runtime -eq "coreclr") {
             if ($NoNative) {
-              _WriteOut "Skipping native image compilation."
+                _WriteOut "Skipping native image compilation."
             }
             else {
-              _WriteOut "Compiling native images for $runtimeFullName to improve startup performance..."
-              Write-Progress -Activity "Installing runtime" "Generating runtime native images" -Id 1
-              Start-Process $CrossGenCommand -Wait -WindowStyle Hidden
-              _WriteOut "Finished native image compilation."
+                _WriteOut "Compiling native images for $runtimeFullName to improve startup performance..."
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@shhsu__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Peter Hsu). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

shhsu:
@davidfowl

glennc:
@shhsu, can you rebase? Is there something we can do for both clr and coreclr? For now I am just hiding both the windows. But I agree that logging the output in a less annoying manner would be better.

shhsu:
I have done some research. I can't find a way to redirect output for Ngen like I did crossgen. Because Ngen requires admin

* I can only use Start-Process -runas ..., I can't use Invoke-Expression (&)
* -runas option conflicts with -NoNewWindow option.

@anurse, can you chime in?

src/dnvm.ps1
```diff
@@ -1044,13 +1044,18 @@ function dnvm-install {
         }
         elseif ($Runtime -eq "coreclr") {
             if ($NoNative) {
-              _WriteOut "Skipping native image compilation."
+                _WriteOut "Skipping native image compilation."
```
glennc: Remove the extra space

shhsu:
Updates?

glennc:
Not sure if there is much you can do for ngen, but no need to block on that we can still merge this one in.

PR #182, opened 2015-03-18
Fix #174 - Allow alias to take a full runtime name instead of just a version

### LINKED ISSUE #174: dnvm alias <alias> <full package name> is broken
I believe that this happened when the `dnvm` was rewritten. `dnvm alias test <full package name>` doesn't work anymore.

I added that feature in https://github.com/aspnet/dnvm/commit/5e71f082970f1079faec1c96862b873c9bbfaff7 but it seems to be gone now.

This break `dev-install` in `DNX` because when it tries to set the alias using the full package name you end up with a runtime named `dnx-clr-win-x86.dnx-clr-win-x64.1.0.0-dev`. This is the exact behavior that `dnvm` had before my change

cc @anurse

anurse:
fix #174 by checking if the input is a full name first

/cc @lodejard @victorhurdugaci


victorhurdugaci:
:shipit:

davidfowl:
Whats the deal with this @anurse ?

anurse:
'twas already merged. Updated the branch to reflect that

PR #194, opened 2015-03-30
Populate the default feed using makefile.shade

pranavkm:


davidfowl:
@pranavkm What's the deal with this?

pranavkm:
The idea is to be able to use the CI to specify the default DNVM feed with it defaulting to the vnext feed.

PR #200, opened 2015-03-30
dnvm install / upgrade should put the installed runtime on path even if ...

Praburaj:
...the runtime is already installed

Fixes: https://github.com/aspnet/dnvm/issues/199

@anurse @glennc


Praburaj:
@muratg

muratg:
:shipit:

PR #203, opened 2015-03-31
Regression causing file not found error when installing coreclr dnx

Praburaj:
This is a regression introduced by https://github.com/aspnet/dnvm/commit/828b00804650a8b77bf324c2f24ca4270b09ef7f.

On install of coreclr xre previously the dnvm use happened before invoking cross gen. With this change I moved this to a common place so that dnvm-use is called irrespective if the dnx already exists or not.

@pranavkm @anurse @muratg


muratg:
:shipit:

PR #208, opened 2015-04-02
Disable progress indicator when script is run in Azure

troydai:
@anurse @glennc @muratg @davidfowl @Eilon


troydai:
http://stackoverflow.com/questions/25678419/how-to-check-if-code-is-running-on-azure-websites/25695126#25695126

anurse:
:shipit:

davidfowl:
Clean

PR #211, opened 2015-04-03
Prepare dnvm.sh for CoreCLR on Linux/OSX

ellismg:
dnvm.sh was unable to handle cases where a runtime package had an os and
architecture included in the name.  For example, using the package name:

  dnx-coreclr-linux-x64.1.0.0-dev

Exposed some bugs. For example, setting an alias for a runtime and then
trying to set that runtime using the alias didn't work.

Fix these issues as well as include the architecture in the output in
dnvm list.  While the CoreCLR team will only be producing x64 of the
runtime in the short term, it will be good to set up the infrastructure
for the day new architectures come online.

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -96,6 +96,14 @@ __dnvm_package_runtime() {
     echo "$runtimeFullName" | sed "s/$_DNVM_RUNTIME_PACKAGE_NAME-\([^.-]*\).*/\1/"
 }
 
+__dnvm_package_arch() {
+    local runtimeFullName="$1"
+    if [[ "$runtimeFullName" =~ $_DNVM_RUNTIME_PACKAGE_NAME-[^-.]*-[^-.]*-[^-.]*\..* ]];
+    then
+        echo "$runtimeFullName" | sed "s/$_DNVM_RUNTIME_PACKAGE_NAME-[^-.]*-[^-.]*-\([^-.]*\)\..*/\1/"
+    fi
+}
+
 __dnvm_update_self() {
     local dnvmFileLocation="$_DNVM_DNVM_DIR/dnvm.sh"
     if [ ! -e $dnvmFileLocation ]; then
@@ -190,7 +198,7 @@ __dnvm_requested_version_or_alias() {
            local runtimeFullName=$(cat "$_DNVM_ALIAS_DIR/$versionOrAlias.alias")
            local pkgName=$(echo $runtimeFullName | sed "s/\([^.]*\).*/\1/")
            local pkgVersion=$(echo $runtimeFullName | sed "s/[^.]*.\(.*\)/\1/")
-           local pkgPlatform=$(echo "$pkgName" | sed "s/$_DNVM_RUNTIME_PACKAGE_NAME-\([^.-]*\).*/\1/")
+           local pkgPlatform=$(echo "$pkgName" | sed "s/$_DNVM_RUNTIME_PACKAGE_NAME-\([^.]*\).*/\1/")
         else
             local pkgVersion=$versionOrAlias
             local pkgPlatform="mono"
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@ellismg__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Matt Ellis). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

ellismg:
@anurse @glennc, mind taking a look?

anurse:
Looks good to me. I assume it still parses `dnx-mono.1.0.0` correctly? (looks like it does)

:shipit: given that.

davidfowl:
Doesn't quite work the way I'd expect:

dnvm list works

dnvm use doesn't:

dnvm use 1.0.0-beta5-11378 -r coreclr

davidfowl:
We need the runtime and arch flags

ellismg:
I will look into adding it.  For now I have just been using the full name instead of using -r and -arch.

davidfowl:
Does that work on powershell? I think they need to be consistent

davidfowl:
And how do you install?

davidfowl:
```
Downloading dnx-mono.dnx-coreclr-win-x86.1.0.0-beta5-11378 from https://www.myget.org/F/aspnetvnext/api/v2
Download: https://www.myget.org/F/aspnetvnext/api/v2/package/dnx-mono/dnx-coreclr-win-x86.1.0.0-beta5-11378
######################################################################## 100.0%
dnx-mono.dnx-coreclr-win-x86.1.0.0-beta5-11378 was not found in repository https://www.myget.org/F/aspnetvnext/api/v2 
Davids-MacBook-Pro:~ dfowler$ 
```

ellismg:
I will play around with that.  I was just making these changes to unblock my simple cases where install is handled by symlinking at dev time.

ellismg:
@davidfowl How do you envision using dnvm use to select a mono based runtime?  It's not even clear to me how to do this in powershell.

FWIW,

```
$ dnvm use dnx-coreclr-linux-x64.1.0.0-dev
Adding /home/matell/.dnx/runtimes/dnx-coreclr-linux-x64.1.0.0-dev/bin to process PATH
```

Works with these changes.

glennc:
We don't really support running Mono on Windows, so the ps1 only gives clr and coreclr.

You shouldn't need to use the full name, `dnvm use 1.0.0-dev` not `dnvm use dnx-coreclr-linux-x64.1.0.0-dev`. To use a coreclr it should be `dnvm use 1.0.0-dev -r coreclr`. To do mono it would be `dnvm use 1.0.0-dev -r mono` for now mono would be the default if you don't select a runtime. This hasn't been required before now because Mono was the only option.

ellismg:
What does it mean to pass an arch when the runtime is mono?  Should we just ignore it?

davidfowl:
Ignore arch on mono

glennc:
I was thinking that when I typed that, and I'm not sure. I think we either ignore it or start producing x64 mono packages :). I am assuming,since we don't have them at the moment, that there is nothing in our packages that changes based on bitness for Mono. So ignore it.

glennc:
We should probably give an error when you pass 32 to the coreclr option until we get a 32 bit dnx.

davidfowl:
We should really add os as well. But lets do that in the next change. I should be able to get any dnx from any dnvm.

glennc:
Yeah, ideally you should be able to bundle for Linux from Windows including a runtime. For example.

ellismg:
@davidfowl  @glennc I added support for -r and -a on use.  I will plumb it across install and the other commands once we actually publish packages.

davidfowl:
I'm almost done reworking the build script

davidfowl:
:shipit:

davidfowl:
Gonna test this out

davidfowl:
The package names will be:

dnx-coreclr-darwin-x64
dnx-coreclr-linux-x64

PR #212, opened 2015-04-04
add run and exec commands

### LINKED ISSUE #128: Add 'exec' and 'run' commands similar to nvm
The node version manager (nvm) has two commands that might be useful in KVM: `exec` and `run`. We should add similar commands:

* dnvm run 1.0.0-beta2 <args> -> Invokes `dnx <args>` for using the 1.0.0-beta2 `dnx`
* dnvm exec 1.0.0-beta2 <cmd> -> Creates a new sub-shell with the 1.0.0-beta2 DNX on the PATH and runs `<cmd>` in it.

Both should take aliases in place of versions as well.

anurse:
Fixes #128 

/cc @glennc @davidfowl 

I believe the PowerShell tests don't pass yet but am not on a Windows box. I'll ensure they pass before merging, but it shouldn't affect the implementation or general structure of the tests (and the sh tests/impl all work) so I figured I'd start the review now.


davidfowl:
AppVeryor says no for some reason

anurse:
Well, it says no because the tests are broken, as I expected. I'll fix them up, but it's purely test issues, not implementation issues.

glennc:
That's what everyone says. Right before the test reveals an implementation issue...

src/dnvm.ps1
```diff
@@ -1180,6 +1180,80 @@ function dnvm-use {
 
 <#
 .SYNOPSIS
+    Locates the dnx.exe for the specified version or alias and executes it, providing the remaining arguments
```
glennc: Seems unfinished. "Locates the dnx.exe for the specified version or alias and executes it, providing the remaining arguments to dnx.exe"?

src/dnvm.ps1
```diff
@@ -1180,6 +1180,80 @@ function dnvm-use {
...
+    The arguments to pass to dnx.exe
+#>
+function dnvm-run {
+    param(
+        [Parameter(Mandatory=$false, Position=0)]
+        [string]$VersionOrAlias,
+        [Parameter(Mandatory=$false, Position=1, ValueFromRemainingArguments=$true)]
+        [object[]]$DnxArguments)
+
+    if([String]::IsNullOrWhiteSpace($VersionOrAlias)) {
```
glennc: Doesn't this mean it is mandatory? Contrary to what the param says...

glennc:
:shipit: Once you get the tests and such working :)

src/dnvm.ps1
anurse: Yeah, we do that elsewhere though. If it's marked as mandatory, we get a not-so-pretty error message and prompt. This is to make the error a little clearer. I'll take a look at it though :)

src/dnvm.ps1
anurse: Yep, will fix.

davidfowl:
:shipit:

PR #219, opened 2015-04-16
update dnvm feed url

anurse:
/cc @glennc @muratg


anurse:
Ignore the travis problems. `release` has a broken travis config.

PR #230, opened 2015-04-24
Fix 229

ellismg:


ellismg:
@glennc PTAL

glennc:
Missed this and included the same change in a PR I just done, I realized this was here as I was doing a PR for the rest of coreclr support :).

PR #233, opened 2015-04-27
Coreclr support

glennc:
Updating dnvm.sh to support CoreCLR.

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -64,14 +64,26 @@ __dnvm_current_os()
 }
 
 __dnvm_find_latest() {
-    local platform="mono"
+    local platform=$1
+    local arch=$2
+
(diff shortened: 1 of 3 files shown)


src/dnvm.sh
```diff
@@ -294,6 +307,7 @@ __dnvm_help() {
     echo "  -p|-persistent    set installed version as default"
     echo "  -f|force          force install. Overwrite existing version of $_DNVM_RUNTIME_SHORT_NAME if already installed"
     echo "  -u|unstable       use unstable feed. Installs the $_DNVM_RUNTIME_SHORT_NAME from the unstable unstable feed"
```
Eilon: Filed https://github.com/aspnet/dnvm/issues/234 for `unstable unstable` on the line above.

PR #236, opened 2015-04-29
add www to nuget url so that it is correct.

glennc:
Fixes: https://github.com/aspnet/dnvm/issues/235


muratg:
Tested. :)

:shipit:

PR #253, opened 2015-05-04
Change dnvm install to move directory instead of moving files.

pranavkm:
Add a guard prior to performing move directory

Fixes #225


pranavkm:
Totally unsure if this is the right fix - but I"m going to try and check this in to see if we can fix the error we see on the CI.

PR #261, opened 2015-05-09
Fixed Get-RuntimeName function if version is passed as alias

### LINKED ISSUE #260: Bug in function Get-RuntimeName
Bug in function Get-RuntimeName

    if(Test-Path $aliasPath) {
        $BaseName = Get-Content $aliasPath

        $Architecture = GetArch $Architecture (Get-PackageArch $BaseName)
        $Runtime = GetRuntime $Runtime (Get-PackageArch $BaseName)
        $Version = Get-PackageVersion $BaseName
    }

I think

        $Runtime = GetRuntime $Runtime (Get-PackageArch $BaseName)

should be

        $Runtime = GetRuntime $Runtime (Get-PackageRuntime $BaseName)

IlyaGrebnov:
This is fixes #260

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -341,8 +341,8 @@ function Get-RuntimeName(
     if(Test-Path $aliasPath) {
         $BaseName = Get-Content $aliasPath
 
-        $Architecture = GetArch $Architecture (Get-PackageArch $BaseName)
-        $Runtime = GetRuntime $Runtime (Get-PackageArch $BaseName)
+        $Architecture = Get-PackageArch $BaseName
+        $Runtime = Get-PackageRuntime $BaseName
         $Version = Get-PackageVersion $BaseName
     }
     


anurse:
Would you be able to add a brief test for this to the powershell tests under `tests/ps1`? It should be able to live within the Alias tests and shouldn't be too difficult to add.

anurse:
Also, this function is used during Install, which should allow you to change the Runtime flavor via a parameter. For example, the following is a valid scenario:

```ps1
# Install a CLR flavor and set it to the test_alias alias
dnvm install 1.0.0 -a test_alias -runtime clr

# Install the same VERSION as test_alias, but the CoreCLR flavor
dnvm install test_alias -runtime coreclr
```

That seems to be the origin of the failing test in AppVeyor.

davidfowl:
@BrennanConroy @anurse @glennc What's the deal with this PR? Is it good to go?

BrennanConroy:
No, appveyor is failing on a test.

IlyaGrebnov:
I think my PR is not correct since test cases are broken :( But there is alternative fix for same issue https://github.com/aspnet/dnvm/pull/248 You should consider merging it instead.

BrennanConroy:
Also, both PR's are doing stuff in a function that no longer exists.

davidfowl:
@BrennanConroy can you advise @IlyaGrebnov based on the latest code?

BrennanConroy:
This got fixed in another PR, see https://github.com/aspnet/dnvm/commit/ad29a7b91a515ee96ff2efdee50f6005f3ac85ec#diff-ddb064ff3406b065f83cfbbd11dc50b3R387

PR #270, opened 2015-05-17
Ensure that the ShowWindow parameter of Get-Help does not cause dnvm-…

### LINKED ISSUE #269: dnvm-help fails on edge case using $PSDefaultParameterValues
## The Problem

I use `$PSDefaultParameterValues` to override `Get-Help` such that the `-ShowWindow` parameter gets defaulted to `$true`. Doing this causes `dnvm` to fail when executing `dnvm-help`.
```
C:\> $PSDefaultParameterValues

Name                           Value
----                           -----
Get-Help:ShowWindow            True
Help:ShowWindow                True
```

```
C:\> dnvm
You must specify a command!
   ___  _  ___   ____  ___
  / _ \/ |/ / | / /  |/  /
 / // /    /| |/ / /|_/ /
/____/_/|_/ |___/_/  /_/
.NET Version Manager v1.0.0-beta5-10376
By Microsoft Open Technologies, Inc.
usage: dnvm <command> [<arguments...>]

Current feed settings:
Default Stable: https://www.nuget.org/api/v2
Default Unstable: https://www.myget.org/F/aspnetvnext/api/v2
Current Stable Override: <none>
Current Unstable Override: <none>

commands:
    alias          ForEach-Object : You cannot call a method on a null-valued expression.
At C:\Users\Jason\.dnx\bin\dnvm.ps1:835 char:13
+             ForEach-Object {
+             ~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [ForEach-Object], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull,Microsoft.PowerShell.Commands.ForEachObjectCommand
```

## The Solution

The problem lies on line 836 of `dnvm.ps1`: `$h = Get-Help $_.Name`.

Modifying line 836 to read `$h = Get-Help $_.Name -ShowWindow:$false` will prevent this edge case from causing `dnvm` to fail.

jrotello:
…help to fail.

Fixes #269

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -833,7 +833,7 @@ function dnvm-help {
         _WriteOut -ForegroundColor $ColorScheme.Help_Header "commands: "
         Get-Command "$CommandPrefix*" | 
             ForEach-Object {
-                $h = Get-Help $_.Name
+                $h = Get-Help $_.Name -ShowWindow:$false
                 $name = $_.Name.Substring($CommandPrefix.Length)
                 if($DeprecatedCommands -notcontains $name) {
                     _WriteOut -NoNewLine "    "
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -751,7 +751,7 @@ function dnvm-help {
             $Script:ExitCodes = $ExitCodes.UnknownCommand
             return
         }
-        $help = Get-Help "dnvm-$Command"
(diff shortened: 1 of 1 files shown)


glennc:
Seems fine. @anurse seem reasonable to you, since this is your bit of the code :)

anurse:
Looks good to me too, merging!

PR #285, opened 2015-06-03
More PSv2.0 work

glennc:
Run tests on PSv2.0. 
Fix some behaviour that was discovered from test runs. 
Change temp path to include runtime name.
Added fallback for k-crossgen command.

/CC @anurse @muratg


muratg:
Looks good to me, but my PowerShell is very weak.  I'll leave giving the squirrel to someone who knows this stuff.

anurse:
:shipit:

PR #290, opened 2015-06-10
Only use ShowWindow when on PowerShell 3 or highter.

glennc:
Update to beta5 now that it is on the release feed, otherwise compiling the test app breaks.
Stop using ShowWindow on PS2


glennc:
@anurse

anurse:
:shipit:

PR #293, opened 2015-06-12
Enabling arm architecture

moozzyk:


glennc:
@moozzyk Can you investigate the appveyor fail? Otherwise change looks good so :shipit:

moozzyk:
@glennc looks like this is the issue with CoreCLR packages we have been trying to fix recently: `C:\projects\dnvm\test\apps\TestApp\Program.cs(5,18): error CS0518: Predefined type 'System.Object' is not defined or imported`

PR #295, opened 2015-06-14
build: Enables FreeBSD support

jasonwilliams200OK:
Changed hardcoded bash shebang to env to support multiple directory
structures (required to build on FreeBSD).

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -208,7 +208,7 @@ __dnvm_unpack() {
 
     #Set shell commands as executable
     find "$runtimeFolder/bin/" -type f \
-        -exec sh -c "head -c 11 {} | grep '/bin/bash' > /dev/null"  \; -print | xargs chmod 775
+        -exec sh -c "head -c 11 {} | grep '/usr/bin/env bash' > /dev/null"  \; -print | xargs chmod 775
 
     #Set dnx to be executable
     chmod 775 "$runtimeFolder/bin/dnx"
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 if test `uname` = Darwin; then
     cachedir=~/Library/Caches/KBuild
--- a/test/sh/chester
+++ b/test/sh/chester
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 VERSION="0.1.0"
 
 # A bunch of helper functions and variables!
--- a/test/sh/run-tests.sh
+++ b/test/sh/run-tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Determine my directory
 SCRIPT_DIR=$(dirname $0)


dougbu:
@jasonwilliams200OK [Wikipedia](https://en.wikipedia.org/wiki/Env) mentions this approach
>  increases the risk that the wrong interpreter is selected because it searches for a match in every directory on the executable search path. It also suffers from the same problem in that the path to the `env` binary may also be different on a per-machine basis.

Is `/usr/bin/env` the correct path in every distribution we care about?

Separately, this change may be necessary but it's likely not sufficient for FreeBSD support. `dnu` (in the DNX repo) hard-codes `/bin/bash` (never `/bin/sh`) and uses that when executing `project.json` scripts as well as when generating script files during `dnu publish` operations.

jasonwilliams200OK:
Admittedly, this is not the best approach, but this is the one we have used in CoreCLR (merged), Roslyn (merged) and CoreFX (under review: https://github.com/dotnet/corefx/pull/2030 see comments).

I have submitted the PRs to aspnet/Universe, aspnet/Home and there already is one opened PR at aspnet/dnx repo.

Copying the FreeBSD team here: @josteink, @ghuntley, @akoeplinger, @janhenke, @ajensenwaud, @saper

josteink:
I can vouch for this being the workaround employed to get support for FreeBSD in the "other" dotnet projects. It's the most portable solution i know to this problem.

dougbu:
:ok: but note this PR and aspnet/dnx#1756 don't quite cover everything yet. I suspect the main gap is in aspnet/dnx#1756 and added a comment there.

src/dnvm.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
anurse: dnvm.sh cannot be executed, it must be `source`-loaded into your current shell, otherwise it can't modify your path. It is intentionally missing a shebang to ensure it isn't accidentally executed. Could you remove this line? It makes sense to have the other ones use the modified shebang, but this file doesn't need it.

dougbu: @anurse I'll add one more commit when I push this tonight.

dougbu:
Sorry @jasonwilliams200OK I am having trouble bringing your branch up-to-date. Could you please rebase your fork and update the `https://github.com/jasonwilliams200OK/dnvm.git dev` server branch?

Believe the following sequence should do it, assuming origin is your server and aspnet ours:
```
git checkout local.branch
git pull --rebase aspnet dev
# deal with conflicts
# squash branches
# test
git push origin local-branch:dev
```

dougbu:
Extra style points awarded if you also restore the first line of `src/dnvm.sh` :smiley_cat:

jasonwilliams200OK:
@dougbu, done :)

dougbu:
Full list of relevant dnvm commits:
- ad8fdd9 (use `/usr/bin/env`)
- 85653d8 (fix bootstrapping issue in above)
- d9713ee (another attempted fix, reverted in next commit)
- 4bc5e73 (pass enough characters for `grep` to succeed)

saper:
Well, there is no need to replace `#!/bin/sh` with env - `/bin/sh` must always be a working shell....

PR #309, opened 2015-06-18
Change the way we invoke commands in order to allow a consistent esca…

glennc:
…ping across PSv2 and PSv3

cc @anurse


anurse:
:shipit: once AppVeyor fixed.

PR #329, opened 2015-06-19
Add architecture and runtime parameters to dnvm run

pranavkm:


pranavkm:
cc @anurse

anurse:
Can you add them to `dnvm exec` as well? Then :shipit: /cc @glennc

Issue #249, opened 2015-05-02
Add override feed help text on Windows


BrennanConroy:
@glennc

glennc:
You mean this text: ``You can control the URL of the stable and unstable channel by setting the DNX_FEED and DNX_UNSTABLE_FEED variables.``?

BrennanConroy:
Yep

glennc:
This has already been done, helping @BrennanConroy out by closing the issues for work he has done...

PR #340, opened 2015-06-28
Help details around which environment vars to set 

avanderhoorn:
I was trying to figure out how to set override feeds and couldn't find any details on this in the console help. This describes the required keys to be set.


davidfowl:
We should probably make the same change to dnvm.sh /cc @glennc

BrennanConroy:
This already exists in the .sh version: https://github.com/aspnet/dnvm/blob/dev/src/dnvm.sh#L271
Issue for this is #249

davidfowl:
Ah!

PR #341, opened 2015-06-29
fix #149 by putting DNX home on the path

### LINKED ISSUE #149: Global commands don't work on Linux/Mac: Add KRE_HOME to path on Linux
Today, we only do that for Windows. 

We need this for `kpm install global` because all commands are installed in the KRE's bin folder.

cc @anurse

anurse:
When dnvm.sh is sourced into the active shell, it will now add the $DNX_HOME/bin value to the PATH.

The PR also contains some whitespace cleanup because my vimrc script that shows trailing whitespace was yelling at me :)

Fixes #149 

/cc @victorhurdugaci @glennc


anurse:
Ping @victorhurdugaci @glennc - Last call before I assume it's good :P

victorhurdugaci:
Looks good. PolyCheck might not like `die` :)

victorhurdugaci:
:shipit:

anurse:
It's never complained before, and it's been there for a while :).

PR #343, opened 2015-07-01
Replace $Host.Version.MajorVersion with $Host.Version.Major

jrotello:
One of the PowerShell host version checks was using the wrong property to evaluate the major version of the host.

A non-existent `$Host.Version.MajorVersion` is being called instead of `$Host.Version.Major`

This pull request updates that check to use the correct property. See other correct usages on lines 755 and 760.

This bug was introduced on 6/10/2015 with commit [21ce03f48a8fc8929957d3ddfba191b203754615](https://github.com/aspnet/dnvm/blame/21ce03f48a8fc8929957d3ddfba191b203754615/src/dnvm.ps1#L841).

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -927,7 +927,7 @@ function dnvm-help {
         _WriteOut -ForegroundColor $ColorScheme.Help_Header "commands: "
         Get-Command "$CommandPrefix*" | 
             ForEach-Object {
-                if($Host.Version.MajorVersion -lt 3) {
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Could you rebase on latest and then I'll merge this in

jrotello:
@BrennanConroy Done.

pranavkm:
@BrennanConroy, can this be merged?

BrennanConroy:
Yes! Thank you

PR #346, opened 2015-07-07
Fix dnvm upgrade in zsh

mattkrea:
Must force overwrite to the alias file in zsh using `>!` instead of `>`.

Tested in both bash and zsh on OS X.

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -623,7 +623,12 @@ dnvm()
             local action="Setting"
             [[ -e "$_DNVM_ALIAS_DIR/$name.alias" ]] && action="Updating"
             echo "$action alias '$name' to '$runtimeFullName'"
-            echo "$runtimeFullName" > "$_DNVM_ALIAS_DIR/$name.alias"
+            
+            if [[ $SHELL == *"zsh"* ]]; then
+                echo "$runtimeFullName" >! "$_DNVM_ALIAS_DIR/$name.alias";
+            else
+                echo "$runtimeFullName" > "$_DNVM_ALIAS_DIR/$name.alias";
+            fi
         ;;
 
         "unalias" )


mattkrea:
This is meant to resolve the following issue using zsh:

```
matt@Matts-MacBook ~ $ dnvm upgrade -u
Determining latest version
Latest version is 1.0.0-beta6-12174
dnx-mono.1.0.0-beta6-12174 already installed.
Adding /Users/matt/.dnx/runtimes/dnx-mono.1.0.0-beta6-12174/bin to process PATH
Updating alias 'default' to 'dnx-mono.1.0.0-beta6-12174'
dnvm:264: file exists: /Users/matt/.dnx/alias/default.alias
```

mattkrea:
Does anyone read these PRs?

muratg:
@mattkrea I don't seem to have this issue (OSX, zsh, dnvm beta 6) Are you sure the file is not open elsewhere? Also, any idea why bash and zsh would behave differently for you?

mattkrea:
It looks like this relates to the [noclobber](http://www.zsh.org/mla/users/2006/msg01186.html) option which I assume by default is turned on with [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) and [prezto](https://github.com/sorin-ionescu/prezto) but I can attempt to confirm that.

anurse:
Looks like this affects bash as well with `noclobber` on. Can you update the PR to handle that case? Unfortunately, in bash the "clobber override" operator is `>|` (that's greater than, followed by pipe :)).

I've verified that this affects `dnvm alias` in both shells on zsh and bash. Can you also add tests to verify this change so we can avoid regressions? In the tests, each test is launched in a separate shell so you should be able to set noclobber as appropriate. Feel free to ask for assistance with writing tests, hopefully you can copy an existing test out of https://github.com/aspnet/dnvm/tree/dev/test/sh/tests/alias

You can run the tests using `.\build verify` from the root of your repo, it will run the tests in **both** `bash` and `zsh`.

mattkrea:
I'll certainly give it a shot

anurse:
Cool. Post a comment when you've either succeeded or given up (which is OK by the way, we can still take the PR) so it shows back up in my notifications :)

mattkrea:
@anurse Still would like to try this but may not get a chance to do this anytime soon with everything going on at work. Just a heads up.

muratg:
@mattkrea will you still be looking at this?

mattkrea:
@muratg Haven't been able to get to it. Sorry. Go ahead and close this out if you'd like.

muratg:
Cool.  Filed https://github.com/aspnet/dnvm/issues/423 to track this issue. Closing this PR as it's going stale.

PR #366, opened 2015-07-24
Modify dnvm to use download url from package feed instead of calculating

### LINKED ISSUE #365: dnvm upgrade does not work if DNX_FEED is pointed at https://www.myget.org/f/aspnetvnext
Repro steps:

```
set DNX_FEED=https://www.myget.org/f/aspnetvnext
dnvm upgrade
```

Expected: Gets the latest and greatest dnx from the feed.
Actual: dnvm fails with a 404.

The problem is that we calculate the download url on the client based off the feed url. In this case, MyGet returns the right response from the root, but with the download url at `/api/v2`: https://www.myget.org/F/aspnetvnext/**api/v2/**package/dnx-clr-win-x86/1.0.0-beta6-12168. The right way to go about this would be to read the content url since this could simply point to a CDN location.

pranavkm:
one on the client.

Fixes #365

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -94,6 +94,15 @@ __dnvm_runtime_bitness_defaults()
     fi
 }
 
+__dnvm_query_feed() {
+    local url=$1
+    xml="$(curl $url 2>/dev/null)"
+    echo $xml | grep \<[a-zA-Z]:Version\>* >> /dev/null || return 1
+    version="$(echo $xml | sed 's/.*<[a-zA-Z]:Version>\([^<]*\).*/\1/')"
+    downloadUrl="$(echo $xml | sed 's/.*<content.*src="\([^"]*\).*/\1/')"
+    echo $version $downloadUrl
+}
+
 __dnvm_find_latest() {
     local platform=$1
     local arch=$2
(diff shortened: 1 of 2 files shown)


pranavkm:
cc @glennc \ @anurse

anurse:
:shipit:

BrennanConroy:
Shell script too?

pranavkm:
@BrennanConroy updated.

src/dnvm.sh
```diff
@@ -88,10 +97,26 @@ __dnvm_find_latest() {
...
+    local platform=$1
+    local arch=$2
+    local version=$3
+
+    if [[ $platform == "mono" ]]; then
+        #dnx-mono
+        local packageId="$_DNVM_RUNTIME_PACKAGE_NAME-$platform"
+    else
+        #dnx-coreclr-linux-x64
+        local packageId="$_DNVM_RUNTIME_PACKAGE_NAME-$platform-$(__dnvm_current_os)-$arch"
```
BrennanConroy: This will need to change based on changes we pushed today

pranavkm: Do you want to go ahead and push your changes to dnvm? I'll rebase on top of it.

BrennanConroy: The changes are already pushed :)

pranavkm:
@BrennanConroy  updated again.

src/dnvm.sh
```diff
@@ -507,14 +532,27 @@ dnvm()
...
+            if [[ "$versionOrAlias" != *.nupkg ]]; then
+                if [[ "$versionOrAlias" == "latest" ]]; then
+                   echo "Determining latest version"
+                   read versionOrAlias downloadUrl < <(__dnvm_find_latest "$runtime" "$arch" "$os")
+                   [[ $? == 1 ]] && echo "Error: Could not find latest version from feed $DNX_ACTIVE_FEED" && return 1
+                   printf "%b\n" "Latest version is ${Cya}$versionOrAlias ${RCol}"
+                else
+                    local runtimeFullName=$(__dnvm_requested_version_or_alias "$versionOrAlias" "$runtime" "$arch" "$os")
+                    local runtimeVersion=$(__dnvm_package_version "$runtimeFullName")
+                    read versionOrAlias downloadUrl < <(__dnvm_find_package "$runtime" "$arch" "$os" "$runtimeVersion")
```
BrennanConroy: $(__dnvm_find_package
new <(__dnvm_find_package
?

pranavkm: That's apparently the right syntax. I tried the script (and it works). For e.g. http://stackoverflow.com/questions/17238608/how-to-return-multiple-variables-from-python-to-bash

BrennanConroy:
Looks good, tested on Mac and Linux?

Issue #368, opened 2015-07-28
dnvm upgrade to beta6 leaves dnu without execute permissions on osx


dennisburton:
After running dnvm upgrade on osx, dnu was correctly placed in the runtimes/[version]/bin directory but did not have execute permissions.

chmod +x dnu functions as a workaround.

BrennanConroy:
What version of dnvm? Try dnvm update-self to get the latest version.

dennisburton:
dnvm was on beta5 at the time of upgrade. Creating a PR for the dnvm.sh file to add the same code for dnu as exists for dnx at ~ line 252

BrennanConroy:
6550ac887e623bf2b32c647814e67969e36cc1e1

Issue #349, opened 2015-07-08
Remove chmod of all bash files


glennc:
We should just do it by name instead of searching for all files that look like they should be executable.

dennisburton:
I tried to tag this in the PR for #369. Probably should have just attached the last commit to this issue

BrennanConroy:
6550ac887e623bf2b32c647814e67969e36cc1e1

PR #369, opened 2015-07-28
Add execute permissions for dnu after unpack

dennisburton:
#368

Add execute permissions for dnu as well as dnx.

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -245,13 +245,14 @@ __dnvm_unpack() {
 
     [ -e "$runtimeFile" ] && rm -f "$runtimeFile"
 
-    #Set shell commands as executable
-    find "$runtimeFolder/bin/" -type f \
-        -exec sh -c "head -c 20 {} | grep '/usr/bin/env bash\|/bin/bash' > /dev/null"  \; -print | xargs -r chmod 775
-
     #Set dnx to be executable
     if [[ -s "$runtimeFolder/bin/dnx" ]]; then
         chmod 775 "$runtimeFolder/bin/dnx"
+    fi
+
+    #Set dnu to be executable
+    if [[ -s "$runtimeFolder/bin/dnu" ]]; then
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Looks good, could you also remove line 258-260? #349

dennisburton:
I'll see if I can dig a bit into why those lines did not work. In theory, if they worked neither dnx nor dnu should have to be modified directly.

dennisburton:
After digging into those lines, it appears  the -r passed to xargs is valid in Linux but not OSX. 
[OSX man](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/xargs.1.html)

Since the goal of -r is to ensure xargs does not run when the input is empty, I did some testing by removing dnu and dnx from the runtimes folder manually before running the script. In this case, chmod is not run since there are no matching shell script files. 

It seems like this PR should:

1. remove the -r option from the xargs call
1. remove lines 251-260 since both dnx and dnu are processed by the new 249-250

@BrennanConroy Is there any testing on this that would validate against different versions of Linux and OSX? I have only been able to test against OSX Yosmite and Ubuntu 14.10.

BrennanConroy:
The issue I linked is about us wanting to remove the entire xargs section and only have the lines for dnx and dnu

Sent from my HTC

----- Reply message -----

dennisburton:
Got it. I'll update and tag both

BrennanConroy:
@glennc @anurse 
Could you squash commits?

anurse:
:shipit: (once commits are squashed). @BrennanConroy can give final sign-off and merge.

dennisburton:
@BrennanConroy do you want me to squash on the PR or are you talking about squashing on merge?

BrennanConroy:
On the PR please :)

BrennanConroy:
Looks great! Thank you for the contribution!

Issue #370, opened 2015-07-28
dnvm upgrade errors on dnvm beta7-10400


henkmollema:
I updated dnvm using `dnvm update-self` to `v1.0.0-beta7-10400`. My feeds are configured as (no overrides):
```
Default Stable: https://www.nuget.org/api/v2
Default Unstable: https://www.myget.org/F/aspnetvnext/api/v2
```
Now when I execute:
```
C:\>dnvm upgrade
```

I get the following error:
> You cannot upgrade to a non-windows runtime. Upgrade will download the latest ve
rsion of the DNX and also set it as your machines default. You cannot set the de
fault DNX to a non-windows version because you cannot use it to run an applicati
on. If you want to install a non-windows DNX to package with your application th
en use 'dnvm install latest -OS:' instead. Install will download the package but
 not set it as your default.
My machine runs on Windows 8.1. It works when I use
```
dnvm upgrade -OS win
```

But why is this necessary? And shouldn't it be documented in https://github.com/aspnet/Announcements/issues/48 for example?

BrennanConroy:
e196ca655a427e151a8ec644591b73bfc51fae19

henkmollema:
Thanks! :+1:

BrennanConroy:
Also, you might need to manually update your dnvm because it looks like update-self was broken too.

henkmollema:
Yeah I notices. Thanks for the heads up.

PR #372, opened 2015-07-28
Upgrade should not error when no OS is provided

BrennanConroy:
#370 
@anurse


avanderhoorn:
Was about to do this as well :D

anurse:
:shipit:

BrennanConroy:
e196ca655a427e151a8ec644591b73bfc51fae19

Issue #254, opened 2015-05-04
dnvm uninstall


jchannon:
how do i delete a version? i can't see any command such as `dnvm uninstall`

davidfowl:
Go to the folder and hit the delete button on the version you want to uninstall.

jchannon:
thats fine but i think if you have an install you should have an uninstall

muratg:
Agreed. @glennc can you have a look when you're back from Ignite?

glennc:
Sure, I can look at it now that I'm back.

moozzyk:
Looks like deleting the runtime folder is not enough. You will leave behind alias files if the runtime you have deleted was aliased.

tugberkugurlu:
I was always **assuming** dnvm had `uninstall`. Taught me a lesson that I will never assume anymore :smile:

jchannon:
You know what they say about assumptions ;)

JasonSoft:
+1

sayedihashimi:
+1 this is needed

muratg:
@BrennanConroy now owns DNVM. I think it would be good to get this in in beta 7.

niemyjski:
+1 I have 4 versions installed It would be nice to be able to uninstall or uninstall on upgrade.

glennc:
The PR is out, just needs to be reviewed and merged.

jchannon:
Link?

glennc:
Linked on this already: https://github.com/aspnet/dnvm/pull/375

muratg:
@BrennanConroy PR for this is out for a long time. Are you going to check this in in Beta 8?

BrennanConroy:
Ya, need to update it since global install went in, will do this week.

BrennanConroy:
6479d250bb15fce17afcc204582849231334f479

Nomad-IC:
wich folder (-s) i must delete for uninstalling?

BrennanConroy:
You can just use dnvm to uninstall or delete the specific runtime in the .dnx/runtimes folder

aligneddev:
I don't see it in dnvm 1.0.0-rc1-update1 clr as an option. It is an "unknown command".

Is that expected?

BrennanConroy:
Are you typing `dnvm uninstall` ?

aligneddev:
[...]

PR #375, opened 2015-07-28
Add dnvm uninstall

BrennanConroy:
#254 
@anurse 

Still need to do powershell.

Currently uninstall is like install, you give it a version, which uses the normal defaults for choosing dnx to uninstall or you can be more specific. An idea was to allow version to be pattern matched, i.e. dnvm uninstall 1.0.0-beta6-*


dougbu:
Should the `uninstall` command
1. Also remove any alias for the removed runtime?
2. Special case removing the active runtime? E.g.
  * Remove the runtime's directory from `$PATH`?
  * Error out and do nothing?
  * Output a warning but do everything?
  * Get all interactive and ask the user what to do?

muratg:
In general it looks good to me. @anurse can you have a look?

muratg:
:watch:

BrennanConroy:
Shell should be good to go @anurse could you take a look?
I'll be fixing powershell half tomorrow

src/dnvm.sh
```diff
@@ -676,6 +683,79 @@ dnvm()
...
+                if [[ $1 == "-r" || $1 == "-runtime" ]]; then
+                    local runtime=$2
+                    shift
+                elif [[ $1 == "-a" || $1 == "-arch" ]]; then
+                    local architecture=$2
+                    shift
+                elif [[ $1 == "-OS" ]]; then
+                    local os=$2
+                    shift
+		elif [[ $1 == "-y" ]]; then
```
anurse: Is this flag documented in the help text above?

test/sh/tests/uninstall/Uninstalling removes dnx.sh
```diff
@@ -0,0 +1,7 @@
+source $COMMON_HELPERS
+source $_DNVM_PATH
+
+$_DNVM_COMMAND_NAME install $_TEST_VERSION
+
+OUTPUT=$($_DNVM_COMMAND_NAME uninstall $_TEST_VERSION)
+echo $OUTPUT | grep "Removed $DNX_USER_HOME/runtimes/dnx-mono.$_TEST_VERSION" || die "expected message was not reported"
```
anurse: Can you test that the DNX was actually removed as well?

anurse:
:watch: @BrennanConroy - Just some small comments

BrennanConroy:
Verbal :shipit:

muratg:
Yes!

anurse:
:+1:

PR #385, opened 2015-08-01
Fix minor incompatibility with zsh

### LINKED ISSUE #384: `dnvm list` fails on zsh
`dnvm list` fails in zsh.
```
$ dnvm list
dnvm:local:311: not valid in this context: dnx-mono.1.0.0-beta7-12290
```

Other commands work fine. It works fine in bash.

natemcmaster:
Fix #384


BrennanConroy:
Verified on Mac?

natemcmaster:
Yup. Works for zsh and bash.

BrennanConroy:
:shipit:

Issue #351, opened 2015-07-10
Show orphaned aliases in output of 'dnvm list'


DamianEdwards:
If I delete the DNX that is currently used by an alias, the alias becomes orphaned but nothing tells me. It would be great if when I run `dnvm list` any orphaned aliases are shown in the table, e.g.:

```
Active Version           Runtime Architecture Location                       Alias
------ -------           ------- ------------ --------                       -----
       1.0.0-beta4       clr     x64          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta4       clr     x86          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta4       coreclr x64          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta4       coreclr x86          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta5       clr     x64          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta5       clr     x86          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta5       coreclr x64          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta5       coreclr x86          C:\Users\dedward\.dnx\runtimes
       1.0.0-beta6-11944 clr     x86                                         default (missing)
       1.0.0-beta6-11944 coreclr x86          C:\Users\dedward\.dnx\runtimes
```

BrennanConroy:
a5012e98285df7d743754edc2185b31d1ef77fe7
2d68778d27225eb04db5994af068560d8e707e83

PR #387, opened 2015-08-06
dnvm list shows orphaned aliases on linux

mholo65:
Implement issue #351 on Linux


Eilon:
@BrennanConroy can you review?

BrennanConroy:
We need to update dnvm.ps1 to show orphaned aliases too.

PR #389, opened 2015-08-07
dnvm list shows orphaned aliases on windows

mholo65:
Implement issue #351 on Windows

@BrennanConroy Here's the PS1 variant for #387

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -359,7 +359,7 @@ function Get-RuntimeAlias {
     if($Aliases -eq $null) {
         _WriteDebug "Scanning for aliases in $AliasesDir"
         if(Test-Path $AliasesDir) {
-            $Aliases = @(Get-ChildItem ($UserHome + "\alias\") | Select-Object @{label='Alias';expression={$_.BaseName}}, @{label='Name';expression={Get-Content $_.FullName }})
+            $Aliases = @(Get-ChildItem ($UserHome + "\alias\") | Select-Object @{label='Alias';expression={$_.BaseName}}, @{label='Name';expression={Get-Content $_.FullName }}, @{label='Orphan';expression={-Not (Test-Path ($RuntimesDir + "\" + (Get-Content $_.FullName)))}})
         } else {
             $Aliases = @()
         }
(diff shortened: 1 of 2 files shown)


mholo65:
Unit Test fails because of this:
"WARNING: column "Location" does not fit into the display and was removed.". Could this be fixed on CI server (appveyor)?

mholo65:
Specifying Width parameter to Out-String, which Format-Table uses to pipe out the table, in function dnvm-list, should do the trick.

Eilon:
@anurse can you review when you're back at the office?

anurse:
Assigning myself in case I don't remember this tomorrow :). @mholo65, if you haven't heard from me in the next day or so, ping this thread! Don't want to miss it and GitHub doesn't seem to have a way to mark notifications as **un**read

mholo65:
@anurse before You can merge this, we need to solve the failing test issue. As I see it, we have several possibilities:

1. Modify test to not use **-Detailed** flag. Drawback: Test can't prove that **Location** is left blank when orphaned alias.
2. Modify shell width (**mode con: cols=4096 lines=4096** or similar) in build.cmd or CI server?
3. Add **-Width** parameter to the **Out-String** command which **Format-Table** pipes out table. Drawback: Will get ugly in small shells where **Location** actually will not fit.
4. Somehow fix that when running tests, shell width is always a large number?

anurse:
Could you modify the test to use the `-PassThru` switch to dnvm list which just returns the raw unformatted PowerShell objects? I'm comfortable with relying on manual testing/customer feedback on the formatted output as long as we know the right data is present in the PowerShell objects.

mholo65:
@anurse  Seems like a good idea. But how do you parse output elegantly since in tests one need to read content from **$__dnvmtest_out**, which is just a string variable.

Running from shell would be as simple as: **dnvm list -passthru | where { $_.Alias -eq 'foobar (missing)' }**

mholo65:
Even worse, there's no content in **$__dnvmtest_out** when I pass **-PassThru** as parameter. Could be because -PassThru bypasses **_WriteOut** function? 

In this case, is it acceptable to skip **__dnvmtest_run** and use **dnvm** command directly?

anurse:
If I understand PowerShell correctly (and I almost definitely don't :stuck_out_tongue_closed_eyes:), the reason the data isn't in $__dnvmtest_out is that _WriteOut is designed to forward the textual output, not the objects. Objects should actually be available as the immediate output of `__dnvmtest_run` itself, so something like `__dnvmtest_run list -passthru | where { $_.Alias -eq 'foobar (missing)' }` might work...

If not, and you can get it to work with `dnvm` directly, go right ahead. `__dnvmtest_run` was just a hacky test wrapper to get output in the right place.

mholo65:
@anurse, got it working :)

mholo65:
@anurse, all OK? Should I start squashing commits?

anurse:
Squash 'em up and I'll merge!

mholo65:
Squashed, please remember to merge #387 ( @BrennanConroy reviewed? ) as well, in order to close #351

PR #393, opened 2015-08-11
dnvm global install

BrennanConroy:
@anurse @glennc 

#414

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -82,6 +82,7 @@ $FullVersion="$ProductVersion-$BuildVersion"
 Set-Variable -Option Constant "CommandName" ([IO.Path]::GetFileNameWithoutExtension($ScriptPath))
 Set-Variable -Option Constant "CommandFriendlyName" ".NET Version Manager"
 Set-Variable -Option Constant "DefaultUserDirectoryName" ".dnx"
+Set-Variable -Option Constant "DefaultGlobalDirectoryName" "Microsoft DNX"
 Set-Variable -Option Constant "OldUserDirectoryNames" @(".kre", ".k")
 Set-Variable -Option Constant "RuntimePackageName" "dnx"
 Set-Variable -Option Constant "DefaultFeed" "https://www.nuget.org/api/v2"
@@ -155,6 +156,7 @@ $DeprecatedCommands = @("unalias")
 # Load Environment variables
 $RuntimeHomes = $env:DNX_HOME
 $UserHome = $env:DNX_USER_HOME
+$GlobalHome = $env:DNX_GLOBAL_HOME
 $ActiveFeed = $env:DNX_FEED
 $ActiveUnstableFeed = $env:DNX_UNSTABLE_FEED
 
@@ -182,9 +184,22 @@ if($CmdPathFile) {
 # Determine where runtimes can exist (RuntimeHomes)
 if(!$RuntimeHomes) {
     # Set up a default value for the runtime home
-    $UnencodedHomes = "%USERPROFILE%\$DefaultUserDirectoryName"
+    $UnencodedHomes = "$env:USERPROFILE\$DefaultUserDirectoryName;$GlobalHome\$DefaultGlobalDirectoryName"
 } else {
     $UnencodedHomes = $RuntimeHomes
+}
+
+# Determine the default global installation directory (GlobalHome)
+if(!$GlobalHome) {
+    if($env:ProgramData) {
+        $GlobalHome = "$env:ProgramData\$DefaultGlobalDirectoryName"
+    } else {
+        $GlobalHome = "$env:AllUsersProfile\$DefaultGlobalDirectoryName"
+    }
+
+    $env:DNX_GLOBAL_HOME="$GlobalHome"
+    $env:DNX_HOME="$env:DNX_HOME;$env:DNX_GLOBAL_HOME"
+    $UnencodedHomes = "$UnencodedHomes;$GlobalHome"
 }
 
 $UnencodedHomes = $UnencodedHomes.Split(";")
@@ -221,6 +236,7 @@ _WriteDebug "Runtime Homes: $RuntimeHomes"
 _WriteDebug "User Home: $UserHome"
 $AliasesDir = Join-Path $UserHome "alias"
 $RuntimesDir = Join-Path $UserHome "runtimes"
+$GlobalRuntimesDir = Join-Path $GlobalHome "runtimes"
 $Aliases = $null
 
 ### Helper Functions
@@ -394,7 +410,7 @@ function Get-RuntimeAliasOrRuntimeInfo(
 }
 
 filter List-Parts {
-    param($aliases)
+    param($aliases, $items)
 
 	$location = ""
 
@@ -420,6 +436,20 @@ filter List-Parts {
     if($parts1[0] -eq "$RuntimePackageName-mono") {
         $parts2 += "linux/osx"
         $parts2 += "x86/x64"
+    }
+
+    $aliasUsed = ""
+    if($items) {
+    $aliasUsed = $items | ForEach-Object {
+        if($_.Architecture -eq $parts2[3] -and $_.Runtime -eq $parts2[1] -and $_.OperatingSystem -eq $parts2[2] -and $_.Version -eq $parts1[1]) {
+            return $true;
+        }
+        return $false;
+    }
+    }
+
+    if($aliasUsed -eq $true) {
+        $fullAlias = ""
     }
 
     return New-Object PSObject -Property @{
@@ -693,7 +723,7 @@ function Get-RuntimePath($runtimeFullName) {
     foreach($RuntimeHome in $RuntimeHomes) {
         $runtimeBin = "$RuntimeHome\runtimes\$runtimeFullName\bin"
         _WriteDebug " Candidate $runtimeBin"
-        if (Test-Path "$runtimeBin") {
+        if (Test-Path $runtimeBin) {
             _WriteDebug " Found in $runtimeBin"
             return $runtimeBin
         }
@@ -976,7 +1006,7 @@ function dnvm-list {
     $RuntimeHomes | ForEach-Object {
         _WriteDebug "Scanning $_ for runtimes..."
         if (Test-Path "$_\runtimes") {
-            $items += Get-ChildItem "$_\runtimes\$RuntimePackageName-*" | List-Parts $aliases
+            $items += Get-ChildItem "$_\runtimes\$RuntimePackageName-*" | List-Parts $aliases $items
         }
     }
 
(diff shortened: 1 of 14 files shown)


src/dnvm.ps1
```diff
@@ -82,6 +82,7 @@ $FullVersion="$ProductVersion-$BuildVersion"
 Set-Variable -Option Constant "CommandName" ([IO.Path]::GetFileNameWithoutExtension($ScriptPath))
 Set-Variable -Option Constant "CommandFriendlyName" ".NET Version Manager"
 Set-Variable -Option Constant "DefaultUserDirectoryName" ".dnx"
+Set-Variable -Option Constant "DefaultGlobalDirectoryName" "Microsoft DNX"
```
anurse: `Microsoft\DNX`

A globally installed runtime should be at: `%ProgramData%\Microsoft\DNX\runtimes\[version]\bin\...`

BrennanConroy: There is already a Microsoft DNX from breadcrumbs, so we decided to use that. We could still change if we want to

anurse: We're synchronizing these folder layouts across a few things over time, so it's best to get everything right now :)

src/dnvm.ps1
```diff
@@ -182,11 +184,25 @@ if($CmdPathFile) {
...
 } else {
     $UnencodedHomes = $RuntimeHomes
 }
 
+# Determine the default global installation directory (GlobalHome)
+if(!$GlobalHome) {
+    
+    $GlobalHome = "$env:ProgramData\$DefaultGlobalDirectoryName"
+    
+    if(!$GlobalHome) {
```
anurse: This will never be true, since $GlobalHome will always have a "\" at the very least.

src/dnvm.ps1
```diff
@@ -182,11 +184,25 @@ if($CmdPathFile) {
 # Determine where runtimes can exist (RuntimeHomes)
 if(!$RuntimeHomes) {
     # Set up a default value for the runtime home
-    $UnencodedHomes = "%USERPROFILE%\$DefaultUserDirectoryName"
+    $UnencodedHomes = "$env:USERPROFILE\$DefaultUserDirectoryName;$GlobalHome\$DefaultGlobalDirectoryName"
```
anurse: The default value for `DNX_HOME` should be based on the values of `DNX_USER_HOME` and `DNX_GLOBAL_HOME`. That way, if I change one of those, I don't have to change `DNX_HOME`

src/dnvm.ps1
```diff
@@ -420,6 +437,20 @@ filter List-Parts {
         $parts2 += "x86/x64"
     }
 
+    $aliasUsed = ""
+    if($items) {
+    $aliasUsed = $items | ForEach-Object {
+        if($_.Architecture -eq $parts2[3] -and $_.Runtime -eq $parts2[1] -and $_.OperatingSystem -eq $parts2[2] -and $_.Version -eq $parts1[1]) {
+            return "true";
```
anurse: `$true` and `$false` are boolean literals in PowerShell. We should use those :)

src/dnvm.ps1
```diff
@@ -1315,7 +1362,7 @@ function dnvm-install {
         Remove-Item $RuntimeFolder -Recurse -Force
     }
 
-    if(Test-Path $RuntimeFolder) {
+    if((Test-Path (Join-Path $RuntimesDir $($runtimeInfo.RuntimeName))) -or (Test-Path (Join-Path $GlobalRuntimesDir $($runtimeInfo.RuntimeName)))) {
         _WriteOut "'$($runtimeInfo.RuntimeName)' is already installed."
```
anurse: Indicate where it is installed so the user can find it. Also, maybe `-Force` should ignore this check? Thoughts @glennc @davidfowl ?

src/dnvm.ps1
anurse: Hmm... Ok... We'll have to consider that. /cc @davidfowl

src/dnvm.ps1
anurse: Also, this check should also scan the "Super-global" installation location (`%ProgramFiles%` on Windows, `/usr/lib` on *nix). DNVM isn't allowed to touch those locations, but it should show them and prevent cross-installation.

glennc: I could live with `<runtimeName> is already installed in <path>`. But it might also be good to explicitly say if the runtime is super global (i.e, you can't uninstall it with dnvm).

src/dnvm.ps1
glennc: I don't like Microsoft DNX either, but it didn't seem bad enough to do the work to change it.

BrennanConroy: Decision?

anurse: Leave it `Microsoft DNX` for now, but we should make sure consider relocating this before RTM to be more consistent with everything else Microsoft :)

BrennanConroy:
@anurse 
Take a look at dnvm.sh please :)

src/dnvm.sh
```diff
@@ -358,15 +372,16 @@ __dnvm_help() {
...
     echo ""
-   printf "%b\n" "${Yel}$_DNVM_COMMAND_NAME upgrade [-f|-force] [-u|-unstable] ${RCol}"
+   printf "%b\n" "${Yel}$_DNVM_COMMAND_NAME upgrade [-f|-force] [-u|-unstable] [-g|-global] ${RCol}"
     echo "  install latest $_DNVM_RUNTIME_SHORT_NAME from feed"
     echo "  adds $_DNVM_RUNTIME_SHORT_NAME bin to path of current command line"
     echo "  set installed version as default"
     echo "  -f|forces         force upgrade. Overwrite existing version of $_DNVM_RUNTIME_SHORT_NAME if already installed"
     echo "  -u|unstable       use unstable feed. Installs the $_DNVM_RUNTIME_SHORT_NAME from the unstable feed"
     echo "  -r|runtime        The runtime flavor to install [clr or coreclr] (default: clr)"
+    echo "  -g|global         Installs the configured global dnx file location (default: /usr/local/lib/)"
```
glennc: How about something like `Installs the latest $_DNVM_RUNTIME_SHORT_NAME in the configured global $_DNVM_RUNTIME_SHORTNAME file location (default: /usr/local/lib current: $DNX_GLOBAL_HOME)`

src/dnvm.sh
```diff
@@ -814,7 +859,7 @@ dnvm()
 }
 
 # Add the home location's bin directory to the path
-export PATH="$DNX_HOME/bin:$PATH"
+export PATH="$DNX_USER_HOME/bin:$PATH"
```
glennc: We will want to support -g for commands as well at some point in the near future. We will want to change this to add the global commands location to the path as well.

glennc: That can probably wait until we do that change though

src/dnvm.sh
anurse: I'd remove the default part, just show `($DNX_GLOBAL_HOME)`. If the user changed the default, they know they changed it :)

anurse:
sh looks ok to me. We're going to need some tests for global install.

src/dnvm.ps1
BrennanConroy: So we should always set DNX_HOME and people shouldn't be changing it themselves

anurse:
If this is needed for beta7, I'm willing to give a hesitant :shipit:, but I'd like to see global install tests (which shouldn't be too hard given we have `DNX_GLOBAL_HOME`) covering the various scenarios: install, install version that already exists, etc.

src/dnvm.sh
```diff
@@ -231,12 +239,28 @@ __dnvm_download() {
         return 1
     fi
 
+    local useSudo=
     mkdir -p "$runtimeFolder" > /dev/null 2>&1
+    if [ ! -d $runtimeFolder ]; then
+	local answer=
+        if [ "$acceptSudo" == "0" ]; then
+            echo "In order to install dnx globally, dnvm will have to temporarily run as root."
+            read -p "You may be prompted for your password via 'sudo' during this process. Is this alright (y/n): " answer
```
anurse: "Is this alright" -> "Is this OK? (y/N)"

(Make the `N` capitalized and ensure that it is the default when nothing is typed)

src/dnvm.sh
```diff
@@ -260,24 +285,24 @@ __dnvm_unpack() {
...
     #Set dnx to be executable
     if [[ -s "$runtimeFolder/bin/dnx" ]]; then
-        chmod 775 "$runtimeFolder/bin/dnx"
+        $useSudo chmod 775 "$runtimeFolder/bin/dnx"
     fi
 
     #Set dnu to be executable
     if [[ -s "$runtimeFolder/bin/dnu" ]]; then
-        chmod 775 "$runtimeFolder/bin/dnu"
+        sudo chmod 775 "$runtimeFolder/bin/dnu"
```
anurse: missing `$useSudo`?

src/dnvm.sh
```diff
@@ -376,6 +410,8 @@ __dnvm_help() {
     echo "  -f|force          force install. Overwrite existing version of $_DNVM_RUNTIME_SHORT_NAME if already installed"
     echo "  -u|unstable       use unstable feed. Installs the $_DNVM_RUNTIME_SHORT_NAME from the unstable feed"
     echo "  -r|runtime        The runtime flavor to install [mono or coreclr] (default: mono)"
+    echo "  -g|global         Installs to the configured global $_DNVM_RUNTIME_SHORT_NAME file location (default: /usr/local/lib/dnx current: $DNX_GLOBAL_HOME)"
+    echo "  -y                Accept use of sudo"
```
anurse: let's match what apt-get says for `-y`: "Assume Yes to all queries and do not prompt"

src/dnvm.sh
BrennanConroy: Nice catch

Sent from my HTC

----- Reply message -----

src/dnvm.sh
```diff
@@ -570,7 +628,7 @@ dnvm()
                 else
                   mkdir -p "$runtimeFolder" > /dev/null 2>&1
                   cp -a "$versionOrAlias" "$runtimeFile"
-                  __dnvm_unpack "$runtimeFile" "$runtimeFolder"
+                  __dnvm_unpack "$runtimeFile" "$runtimeFolder" sudo
```
anurse: Missing `$useSudo`?

anurse:
sh changes look good to me. I think we might want a way for people to force install even if the runtime is installed in a different location. Imagine trying to "promote" a runtime you've already installed locally to the global store... right now we don't have a way to do that other than manually copying the directory.

BrennanConroy:
Take a look at the windows part?

anurse:
:shipit:

PR #402, opened 2015-08-20
Which dnvm

mishfit:
It's silly but dnvm went from brew install to a simple curl call and might throw off those that installed using brew previously (or perhaps it will only ever throw me off and no one else)

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -201,7 +201,12 @@ __dnvm_update_self() {
     if [ ! -e $dnvmFileLocation ]; then
         local formattedDnvmFileLocation=`(echo $dnvmFileLocation | sed s=$HOME=~=g)`
         local formattedDnvmHome=`(echo $_DNVM_DNVM_DIR | sed s=$HOME=~=g)`
-        printf "%b\n" "${Red}$formattedDnvmFileLocation doesn't exist. This command assumes you have installed dnvm in the usual location and are trying to update it. If you want to use update-self then dnvm.sh should be sourced from $formattedDnvmHome ${RCol}"
+        local bashSourceLocation=${BASH_SOURCE}
+        local scriptLocation=$bashSourceLocation
+        if [ -z "${bashSourceLocation}" ]; then
+          local scriptLocation=${(%):-%x}
+        fi
+        printf "%b\n" "${Red}$formattedDnvmFileLocation doesn't exist. This command assumes you have installed dnvm in the usual location and are trying to update it. If you want to use update-self then dnvm.sh should be sourced from $formattedDnvmHome. dnvm is currently sourced from $scriptLocation ${RCol}"
         return 1
     fi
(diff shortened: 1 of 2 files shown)


mishfit:
@pranavkm can you take a(nother) look?

mishfit:
Since Window's has `where` I figured I didn't need to touch dnvm.cmd or *.ps1

src/dnvm.sh
```diff
@@ -194,7 +194,7 @@ __dnvm_update_self() {
     if [ ! -e $dnvmFileLocation ]; then
         local formattedDnvmFileLocation=`(echo $dnvmFileLocation | sed s=$HOME=~=g)`
         local formattedDnvmHome=`(echo $_DNVM_DNVM_DIR | sed s=$HOME=~=g)`
-        printf "%b\n" "${Red}$formattedDnvmFileLocation doesn't exist. This command assumes you have installed dnvm in the usual location and are trying to update it. If you want to use update-self then dnvm.sh should be sourced from $formattedDnvmHome ${RCol}"
+        printf "%b\n" "${Red}$formattedDnvmFileLocation doesn't exist. This command assumes you have installed dnvm in the usual location and are trying to update it. If you want to use update-self then dnvm.sh should be sourced from $formattedDnvmHome. dnvm is currently sourced from $BASH_SOURCE ${RCol}"
```
pranavkm: We currently support zsh along with Bash. Would this need to be written like http://stackoverflow.com/a/15056664/235591 to make sure the sourced location works in both these shells?

mishfit: Yep, it would need that...

mishfit:
@pranavkm I made the change and *sort of* made a test (which doesn't currently assert anything)

1. it pushes dnvm.sh to a new "non-standard" directory
2. calls source on the new location
3. calls `dnvm update-self`
  * this is where I would need to parse the output and determine that a location was supplied

On another note, here's what I see when I'm in either bash or zsh:
which-dnvm-zsh-v-bash

You can tell that it's working for bash but how zsh determines what constitutes the called script appears to be funny and doesn't provide useful info

BrennanConroy:
$0 is the function name, look at this: http://stackoverflow.com/questions/9901210/bash-source0-equivalent-in-zsh

pranavkm:
`%x` according to http://stackoverflow.com/questions/9901210/bash-source0-equivalent-in-zsh#comment45012039_23259585?

mishfit:
Okay, now the verbiage is clearer in both bash and zsh:
sourcing

Would love suggestions on testing it (I guess I could export a variable and check it but it wouldn't be very clean to leave it hanging around just for testing)

pranavkm:
I don't think we have existing tests for the current verbiage. Might be ok to check this one in as is.
cc @glennc , does that sound alright?

BrennanConroy:
Rebase please

BrennanConroy:
Could you squash into a single commit please?

PR #415, opened 2015-09-01
Prevent DNX_HOME/bin to be added repeatedly to PATH when sourcing dnvm.sh

### LINKED ISSUE #412: [Linux] `. ~/.dnx/dnvm/dnvm.sh` repeatedly adds `~/.dnx/bin` to `$PATH`
`~/.dnx/bin` does not exist by default. However sourcing the `dnvm.sh` file adds it to the `$PATH` every time. I noticed my `$PATH` had a couple of these entries after sourcing the file twice.

mholo65:
Fixes #412 

I also needed to update the dnx-mono nupkg used in tests to a newer version. Didn't seem to find old beta6-12208 on feed anymore. Added this as separate commit.

//cc @dougbu @BrennanConroy @anurse

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -813,8 +813,8 @@ dnvm()
     return 0
 }
 
-# Add the home location's bin directory to the path
-export PATH="$DNX_HOME/bin:$PATH"
+# Add the home location's bin directory to the path if it doesn't exist
+[[ ":$PATH:" != *":$DNX_HOME/bin:"* ]] && export PATH="$DNX_HOME/bin:$PATH"
 
 # Generate the command function using the constant defined above.
 $_DNVM_COMMAND_NAME list default >/dev/null && $_DNVM_COMMAND_NAME use default >/dev/null || true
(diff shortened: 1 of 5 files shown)


anurse:
:shipit: @BrennanConroy

src/dnvm.sh
```diff
@@ -813,8 +813,8 @@ dnvm()
     return 0
 }
 
-# Add the home location's bin directory to the path
-export PATH="$DNX_HOME/bin:$PATH"
+# Add the home location's bin directory to the path if it doesn't exist
+[[ ":$PATH:" != *":$DNX_HOME/bin:"* ]] && export PATH="$DNX_HOME/bin:$PATH"
```
dougbu: Why do this at all? Nothing creates or uses the `$DNX_HOME/bin` directory.

BrennanConroy: $DNX_HOME/bin is the home of global commands and possibly dnvm if we decide to move it there like on Windows.

PR #416, opened 2015-09-01
Fixed dnvm list x-plat inconsistency. 'Arch' renamed to 'Architecture…

### LINKED ISSUE #390: dnvm list inconsistent between Win and *nix/OSX
Column order is different when using -detailed flag.

dnvm.sh uses:
```
"Active" "Version" "Runtime" "Architecture" "OperatingSystem" "Alias" "Location"
```

and dnvm.ps1 uses;
```
"Active" "Version" "Runtime" "Arch" "OperatingSystem" "Location" "Alias"
```

As you can see from the above, Arch is used on *nix/OSX while Architecture is used on WIn

mholo65:
…', 'Location' moved as last column when using -detailed flag

Fixes #390 

Tests will fail since beta6-12208 can't be found anymore. Merge #415 prior to this to fix that issue.

//cc @BrennanConroy @anurse

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -763,13 +763,33 @@ dnvm()
             fi
 
             if [[ $2 == "-detailed" ]]; then
-                local formatString="%-6s %-20s %-7s %-4s %-15s %-20s %s\n"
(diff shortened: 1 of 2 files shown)


anurse:
:shipit: after #415 merged

PR #426, opened 2015-09-06
Check if directory exist before executing find

### LINKED ISSUE #425: dnvm list outputs "find: `/usr/local/lib/dnx/runtimes': No such file or directory"
After global install was added, dnvm list now outputs the message "find: `/usr/local/lib/dnx/runtimes': No such file or directory".

This error message is seen everytime I open up the terminal, since sourcing dnvm.sh automatically runs ```dnvm list``` and ```dnvm use```

```dnvm list``` should check if directory exist before executing find.

mholo65:
Fixes #425 by checking if directory exist before executing find. If e.g. global install dir is not present, one will get "find: `/usr/local/lib/dnx/runtimes': No such file or directory" error message everytime opening a terminal or running dnvm list.

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -762,7 +762,7 @@ dnvm()
             local name="$1"
 
             if [[ $# == 1 ]]; then
-                [[ ! -e "$_DNVM_ALIAS_DIR/$name.alias" ]] && echo "There is no alias called '$name'" && return
+                [[ ! -e "$_DNVM_ALIAS_DIR/$name.alias" ]] && echo "There is no alias called '$name'" && return 1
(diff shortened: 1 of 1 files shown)


muratg:
cc: @BrennanConroy

BrennanConroy:
This looks good, but as you pointed out in the linked issue dnvm list default isn't valid, could you change it to dnvm alias default? (Very last line of dnvm.sh)

mholo65:
@BrennanConroy it's changed now.

To make the change make sense, I also changed ```dnvm alias``` to return non-zero if requested alias was not found.

BrennanConroy:
:+1:

PR #427, opened 2015-09-06
dnvm alias allows updating existing alias if noclobber is set

### LINKED ISSUE #423: dnvm upgrade fails if noclobber is set (Linux/OSX)
Overwrite operator `>` fails both on Bash and ZSH.  Need to use `>!` on ZSH and `>|` in Bash instead.

See: https://github.com/aspnet/dnvm/pull/346

mholo65:
Fixes #423 

@anurse would it be beneficial to run all tests with ```noclobber``` option?

Accoring to the [Unix Power Tools Book](https://books.google.fi/books?id=Xk6THylQxRUC&lpg=SA43-PA6&dq=safe%20I%2FO%20redirection%20with%20noclobber&pg=SA43-PA6#v=onepage&q=safe%20I/O%20redirection%20with%20noclobber&f=false) the Z shell understands both ```!```and ```|```, therefore I'm not using any shell checking in ```dnvm.sh```

Regarding checking current shell, unlike the original PR, I'm not checking ```$SHELL``` variable as this will only return default shell, therefore checking ```$BASH_VERSION``` or ```$ZSH_VERSION``` would be more safe.

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -793,7 +793,7 @@ dnvm()
             local action="Setting"
             [[ -e "$_DNVM_ALIAS_DIR/$name.alias" ]] && action="Updating"
             echo "$action alias '$name' to '$runtimeFullName'"
-            echo "$runtimeFullName" > "$_DNVM_ALIAS_DIR/$name.alias"
+            echo "$runtimeFullName" >| "$_DNVM_ALIAS_DIR/$name.alias"
         ;;
 
         "unalias" )
--- a/test/sh/tests/alias/Aliasing replaces existing alias.sh
+++ b/test/sh/tests/alias/Aliasing replaces existing alias.sh
@@ -1,6 +1,15 @@
 source $COMMON_HELPERS
 source $_DNVM_PATH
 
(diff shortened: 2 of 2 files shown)


anurse:
We need to get automated tests using dash, since that's the default shell in Ubuntu...

Until then, do you have access to a Ubuntu machine (or other machine with dash installed) to give it a quick try there and make sure `>|` works in dash?

Otherwise, looks good to me! Thanks! /cc @BrennanConroy

mholo65:
According to [manual](http://manpages.ubuntu.com/manpages/precise/man1/dash.1.html), dash uses ```>|```
```
[n]> file   Redirect standard output (or n) to file.
[n]>| file  Same, but override the -C option.
```

Also verified on my own computer:
```sh
$ touch foobar
$ dash -C -c "echo 'foobar' > foobar"
dash: 1: cannot create foobar: File exists
$ dash -C -c "echo 'foobar' >| foobar"
$ cat foobar
foobar
```

muratg:
Looks good to me.  @BrennanConroy, could you take it in?

Issue #428, opened 2015-09-09
`dnvm run ...` and `dnvm exec` do not set exit code correctly


dougbu:
`dnvm run` and `dnvm exec` do not pass a non-`0` exit code from `dnx` back to the shell. (Side note: This and the `_k-test.shade` / `_k.shade` approach means DNX Core tests on Windows can fail without ending the build.)

Repro steps
1. create a simple Console application that always fails
2. `dnu restore`
3. `dnx run`
4. echo %ErrorLevel%`
  - observe expected `1`
5. `dnvm run default run` or `dnvm exec default dnx run`
6. `echo %ErrorLevel%`
  - observe *unexpected* `0`

Example program
``` c#
namespace ExitCode
{
    public class Program
    {
        public int Main(string[] args)
        {
            return 1;
        }
    }
}
```

muratg:
@dougbu Is this a regression?

dougbu:
I don't think this ever worked correctly. Not sure why we changed KoreBuild as if it did. @BrennanConroy ??

BrennanConroy:
4f0b0994091d934de8ea316d31c4a51a56c1932c

PR #434, opened 2015-09-10
Return exit code from dnx in dnvm exec and run

BrennanConroy:
@dougbu @anurse 
#428

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -1610,6 +1610,7 @@ function dnvm-run {
     }
     _WriteDebug "> $dnxExe $DnxArguments"
     & $dnxExe @DnxArguments
+    $Script:ExitCode = $LASTEXITCODE
 }
 
 <#
@@ -1652,6 +1653,7 @@ function dnvm-exec {
         & $Command @Arguments
     } finally {
         $env:PATH = $oldPath
+        $Script:ExitCode = $LASTEXITCODE
     }
 }
 
--- a/test/ps1/tests/RunAndExec.Tests.ps1
+++ b/test/ps1/tests/RunAndExec.Tests.ps1
@@ -6,7 +6,7 @@ Describe "run" -Tag "run" {
             __dnvmtest_run use none | Out-Null
             { Get-Command dnx -ErrorAction Stop } | Should Throw
             (__dnvmtest_run run $TestRuntimeVersion | Select-String -SimpleMatch "Microsoft .NET Execution environment CLR-x86-$TestRuntimeVersion") | Should Be $true
-            $__dnvmtest_exit | Should Be 0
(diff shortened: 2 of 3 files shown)


src/dnvm.ps1
```diff
@@ -1652,6 +1653,7 @@ function dnvm-exec {
         & $Command @Arguments
     } finally {
         $env:PATH = $oldPath
+        $Script:ExitCode = $LASTEXITCODE
```
dougbu: Suggest moving this up one line just to make things super clear.

dougbu: Also, `$LastExitCode` doesn't work with `dnx` because PowerShell doesn't seem to think `dnx.exe` is a Windows command. I'm not sure how to get the correct code without invoking `cmd` and somehow returning `%ErrorLevel%` to PowerShell. `$?` does correctly indicate `dnx.exe` failed however -- just doesn't provide the exact code.

For example
```
...\ExitCode> dnx run
...\ExitCode> $LASTEXITCODE
0
...\ExitCode> cmd /c dnx run
...\ExitCode> $LASTEXITCODE
0
...\ExitCode> cmd
Microsoft Windows [Version 6.3.9600]
(c) 2013 Microsoft Corporation. All rights reserved.
...\ExitCode>dnx run

...\ExitCode>echo %ERRORLEVEL%
1
```

test/ps1/tests/RunAndExec.Tests.ps1
```diff
@@ -6,7 +6,7 @@ Describe "run" -Tag "run" {
             __dnvmtest_run use none | Out-Null
             { Get-Command dnx -ErrorAction Stop } | Should Throw
             (__dnvmtest_run run $TestRuntimeVersion | Select-String -SimpleMatch "Microsoft .NET Execution environment CLR-x86-$TestRuntimeVersion") | Should Be $true
-            $__dnvmtest_exit | Should Be 0
+            $__dnvmtest_exit | Should Be 2
```
dougbu: do we have any tests of `dnvm.sh`?

dougbu:
:watch:

src/dnvm.ps1
BrennanConroy: Do you have some strange settings because it works fine with my powershell

dougbu: Found I had a buried bug in my Prompt function -- looked like it was doing some work then setting `$LastExitCode` correctly but it was just a mirage. Fixed that and I don't see an issue with your PR :smiley_cat:

dougbu:
:shipit:

BrennanConroy:
4f0b0994091d934de8ea316d31c4a51a56c1932c

Issue #422, opened 2015-09-04
Bug - VSO Build Prebuild.ps1 CoreClr


Eneuman:
Using the latest dnvm seems to break VSO builds.
(To reprocude this issue, follow the information in this post: https://msdn.microsoft.com/en-us/Library/vs/alm/Build/azure/deploy-aspnet5 )

My Prebuild.ps1 has this block:

echo "Installing clr"
& $env:USERPROFILE\.dnx\bin\dnvm install $dnxVersion -r clr -Persistent
echo "Installing coreclr"
& $env:USERPROFILE\.dnx\bin\dnvm install $dnxVersion -r coreclr 

(The rest of it has not been changed)
Installing just clr works great and the build finish successfully.
But then I try to also install coreclr I get this error message:

Installing clr
Determining latest version
Downloading dnx-clr-win-x86.1.0.0-beta7 from https://www.nuget.org/api/v2
Installing to C:\Users\buildguest\.dnx\runtimes\dnx-clr-win-x86.1.0.0-beta7
Adding C:\Users\buildguest\.dnx\runtimes\dnx-clr-win-x86.1.0.0-beta7\bin to process PATH
Adding C:\Users\buildguest\.dnx\runtimes\dnx-clr-win-x86.1.0.0-beta7\bin to user PATH
Installing coreclr
Join-Path : Cannot bind argument to parameter 'Path' because it is an empty string.
At C:\Users\buildguest\.dnx\bin\dnvm.ps1:207 char:59
+ $RuntimeDirs = $RuntimeHomes | ForEach-Object { Join-Path $_ "runtimes" }
+ ~~
+ CategoryInfo : InvalidData: (:) [Join-Path], ParameterBindingValidationException
+ FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed,Microsoft.PowerShell.Commands.Join 
PathCommand

Determining latest version
Downloading dnx-coreclr-win-x64.1.0.0-beta7 from https://www.nuget.org/api/v2
Installing to C:\Users\buildguest\.dnx\runtimes\dnx-coreclr-win-x64.1.0.0-beta7
Cannot find dnx-coreclr-win-x64.1.0.0-beta7, do you need to run 'dnvm install 1.0.0-beta7'?
At C:\Users\buildguest\.dnx\bin\dnvm.ps1:1549 char:9
+ throw "Cannot find $runtimeFullName, do you need to run '$CommandName in ...
[...]

Issue #391, opened 2015-08-10
dnvm update-self error: Cannot bind argument to parameter Path


per-samuelsson:
Since some time back, `dnvm update-self` gives me this error.

```
%home%>dnvm update-self
Updating dnvm from https://raw.githubusercontent.com/aspnet/Home/dev/dnvm.ps1
Invoke-Command : Cannot bind argument to parameter 'Path' because it is null.
At C:\Users\Per\.dnx\bin\dnvm.ps1:1479 char:23
+         Invoke-Command <<<<  ([ScriptBlock]::Create("dnvm-$cmd $cmdargs"))
    + CategoryInfo          : InvalidData: (:) [Invoke-Command], ParameterBind 
   ingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,M 
   icrosoft.PowerShell.Commands.InvokeCommandCommand
```

I installed dnvm originally using the [standalone installation](http://docs.asp.net/en/latest/getting-started/installing-on-windows.html#install-the-net-version-manager-dnvm), and I'm on a **Win7 x64** environment. I'm running the above in the **standard command-prompt, not elevated** (but running it as an admin shows the same thing).

I've configured NuGet to be able to grab bits from the unstable feeds, and bringing down runtimes works fine. The only problem I have is the above.

Here's some dnvm info:

```
.NET Version Manager v1.0.0-beta6-10394
Current feed settings:
Default Stable: https://www.nuget.org/api/v2
Default Unstable: https://www.myget.org/F/aspnetvnext/api/v2
Current Stable Override: <none>
Current Unstable Override: <none>
```

Any help/pointers appreciated!

BrennanConroy:
@per-samuelsson It looks like you were unfortunate enough to grab the version ```dnvm update-self``` broke. Sorry about that! All you need to do is run the installation again using the link you provided and you should get a newer version with a working ```dnvm update-self```

glennc:
@per-samuelsson Closing this, if updating DNVM doesn't fix it then reopen it.

per-samuelsson:
@BrennanConroy Thanks for helping out!

> Sorry about that!

No worries at all. Theres no panic zone or anything.

> All you need to do is run the installation again using the link you provided and you should get a newer version with a working `dnvm update-self`

[...]

PR #441, opened 2015-09-17
Fix for dnvm update-self on PowerShell 2.0

hubuk:
This changeset consist of two fixes for dnvm update-self when executed on PowerShell 2.0:
- PSScriptRoot variable is not available in PowerShell 2.0. This change sets a value for this variable if it is not set or empty.
- When DNX_HOME environmental variable has not been set yet the script will put a semicolon at the beginning of the value. This change removes empty entries when splitting the string by ';'

Fix for:
#391
#422

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -846,6 +846,14 @@ function Is-Elevated() {
     return $user.IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator")
 }
 
+function Get-ScriptRoot() {
+    if (!$PSVersionTable.PSVersion.Major -ge 3) {
+        return $PSScriptRoot
+    }
+
+    return Split-Path $script:MyInvocation.MyCommand.Path -Parent
+}
+
 ### Commands
 
 <#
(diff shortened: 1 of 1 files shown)


BrennanConroy:
#422 was fixed, could you rebase and see why the tests didn't pass?

hubuk:
I have removed my attempt to fix #422. Fix for #391 has been slightly changed (renamed a function). Build succeeded, but I do not know why it failed before.

muratg:
@glennc Can you have a look?

muratg:
@glennc if you won't be able to look let me know.

glennc:
Looks reasonable to me. I say :shipit: unless @anurse or someone thinks otherwise :). Will give a little time and then merge.

anurse:
:+1: I wonder it it would be possible for us to run our tests against PowerShell 2.0. I believe PowerShell lets you specify a version to reduce the compatibility mode of the shell.

glennc:
Almost 100% certain we already do that. But it doesn't catch everything.

glennc:
https://github.com/aspnet/dnvm/blob/dev/test/ps1/Run-Tests.ps1#L94

glennc:
Thanks for the PR :)

PR #444, opened 2015-09-17
fixed dnx-mono package version

devel0:
Thanks to mholo65 suggestion I updated the run-test.sh with the next avail version of dnx-mono 15594 instead of 15530, updated the hash using sha1sum in the script and now it works for me.


mholo65:
Wouldn't it be better to also change ```DNX_FEED``` to ```https://www.nuget.org/api/v2``` and change ```_TEST_VERSION``` to ```1.0.0-beta7``` as this will break in a couple of days (once ```1.0.0-beta8-15594``` gets deleted from MyGet)?

//cc @anurse @brennanconroy

BrennanConroy:
I agree with @mholo65 changing to beta7 on NuGet will guarantee the nupkg being around so we don't need to keep updating this

devel0:
I agree for avoid continuos changing,
but if is important to test dnvm with latest dnx-mono package (as this is a dev branch of dnvm)
maybe useful to enable the test script to get the latest dnx-mono in a dynamic way by querying the nuget repo for unstable package and getting the latest automatically.

mholo65:
AFAIK version of the package is not important. It's more important to get a working package. By always downloading latest, we can't assure that package is working as we can't check the hash.

BTW, it seems that ps1 is also broken due to same reason.

devel0:
Yes the ps1 have the same problem as the sh,
so definitively fixing that at beta7 to now would be a solution.
Instead, about the hash I expected one could get the hash for a specific package from an official https server ( that way you can trust identity of the server and data integrity that came from it ). I'm wrong ?
Sorry I'm not experienced with nuget package system, I'm just curious about to learn something more.

mholo65:
It's possible to obtain hash from NuGet server. Have a look at  [this](https://www.myget.org/F/aspnetvnext/api/v2/GetUpdates()?packageIds=%27dnx-mono%27&versions=%270.0%27&includePrerelease=true&includeAllVersions=false)

By parsing that XML, You get latest version, hash (sha512), and download link.

However, doing this requires some work, and since tests rather needs working version than latest version, I think that it's not worth it.

devel0:
Thanks again mholo65 I understood,
meanwhile I fixed the ps1 as discussed using 1.0.0-beta7

While the shell version using 1.0.0-beta7 gets different behavior : it will fail on "test/sh/tests/Aliases can be listed.sh".
Here seems it uses $_TEST_VERSION that would be '1.0.0-beta7' as full package name while it should use $_NUPKG_FILE in my opinion but I'm sure about that cause with the current beta8 it works without need to change anything on the rest.

So in the Unix version someone else that know how script organized should fix that to work on 1.0.0-beta7.

mholo65:
Hmmm, did you also change ```DNX_FEED```? Tested and had no problem with changing to ```1.0.0-beta7``` on ```nuget.org```.

Anyway, here's my diff:
```diff
diff --git a/test/sh/run-tests.sh b/test/sh/run-tests.sh
index f92adda..db4248f 100755
--- a/test/sh/run-tests.sh
+++ b/test/sh/run-tests.sh
@@ -20,14 +20,14 @@ source $COMMON_HELPERS
 [ -z "$TEST_SHELLS" ]       && export TEST_SHELLS="bash zsh"
 [ -z "$TEST_DIR" ]          && export TEST_DIR="$SCRIPT_DIR/tests"
 [ -z "$CHESTER" ]           && export CHESTER="$SCRIPT_DIR/chester"
-[ -z "$DNX_FEED" ]          && export DNX_FEED="https://www.myget.org/F/aspnetvnext/api/v2" # doesn't really matter what the feed is, just that it is a feed
+[ -z "$DNX_FEED" ]          && export DNX_FEED="https://www.nuget.org/api/v2" # doesn't really matter what the feed is, just that it is a feed
 [ -z "$TEST_APPS_DIR" ]     && export TEST_APPS_DIR="$REPO_ROOT/test/apps"
 
 export DNX_FEED
 
 # This is a DNX to use for testing various commands. It doesn't matter what version it is
-[ -z "$_TEST_VERSION" ]     && export _TEST_VERSION="1.0.0-beta8-15530"
-[ -z "$_NUPKG_HASH" ]       && export _NUPKG_HASH='1abc9039fbf10ec3eebc9af216be103ffd1ccb7b'
+[ -z "$_TEST_VERSION" ]     && export _TEST_VERSION="1.0.0-beta7"
+[ -z "$_NUPKG_HASH" ]       && export _NUPKG_HASH='234e543292254e18fe57c6f77c54b281d9d600b7'
 [ -z "$_NUPKG_URL" ]        && export _NUPKG_URL="$DNX_FEED/package/$_DNVM_RUNTIME_PACKAGE_NAME-mono/$_TEST_VERSION"
 [ -z "$_NUPKG_NAME" ]       && export _NUPKG_NAME="$_DNVM_RUNTIME_PACKAGE_NAME-mono.$_TEST_VERSION"
 [ -z "$_NUPKG_FILE" ]       && export _NUPKG_FILE="$TEST_WORK_DIR/${_NUPKG_NAME}.nupkg"
```

devel0:
I tried to apply the patch starting from the dev branch too but with another problem now, you not got this ? maybe some config in my build host, lets see what reports continuos-integration servers.

```
error:       Warning: DownloadPackageAsync: https://www.myget.org/F/aspnetlitedev/api/v2/package/System.Console/4.0.0-beta-23302
error:         The requested feature is not implemented.
error:         GET https://www.myget.org/F/aspnetlitedev/api/v2/package/System.Console/4.0.0-beta-23302
```

mholo65:
Didn't experience any of those errors, seems like your machine have some issues as also CI servers are passing. Anyhow, this branch does now have conflicts (because of 416db7f). Resolve the conflicts and squash commits.

@BrennanConroy it would be nice to get this one merged so we don't need to keep updating this.

devel0:
did it, thx for directions

BrennanConroy:
Awesome, now we shouldn't have to change this for a long time :)

PR #446, opened 2015-09-18
Reorder `dnvm.ps1` blocks to avoid read-before-write issues

dougbu:
- #422
- previously used `$env:DNX_HOME` or `$env:DNX_GLOBAL_HOME` when undefined
- also ignores `$env:DNX_HOME` values such as ";C:\ProgramData\Microsoft DNX"

nits:
- use constants a bit more consistently
- remove trailing whitespace

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -109,8 +109,8 @@ Set-Variable -Option Constant "DNVMUpgradeUrl" "https://raw.githubusercontent.co
 Set-Variable -Option Constant "AsciiArt" @"
    ___  _  ___   ____  ___
   / _ \/ |/ / | / /  |/  /
- / // /    /| |/ / /|_/ / 
-/____/_/|_/ |___/_/  /_/  
+ / // /    /| |/ / /|_/ /
+/____/_/|_/ |___/_/  /_/
 "@
 
 $ExitCodes = @{
@@ -154,11 +154,11 @@ if($__TeeTo) {
 $DeprecatedCommands = @("unalias")
 
 # Load Environment variables
-$RuntimeHomes = $env:DNX_HOME
+$RuntimeHomes = $(if (Test-Path "env:\$HomeEnvVar") {Get-Content "env:\$HomeEnvVar"})
 $UserHome = $env:DNX_USER_HOME
 $GlobalHome = $env:DNX_GLOBAL_HOME
-$ActiveFeed = $env:DNX_FEED
-$ActiveUnstableFeed = $env:DNX_UNSTABLE_FEED
+$ActiveFeed = $(if (Test-Path "env:\$DefaultFeedKey") {Get-Content "env:\$DefaultFeedKey"})
+$ActiveUnstableFeed = $(if (Test-Path "env:\$DefaultUnstableFeedKey") {Get-Content "env:\$DefaultUnstableFeedKey"})
 
 # Default Exit Code
 $Script:ExitCode = $ExitCodes.Success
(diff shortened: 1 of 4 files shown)


src/dnvm.ps1
```diff
@@ -181,13 +181,32 @@ if($CmdPathFile) {
...
+      $pf = $env:ProgramFiles
+      if(Test-Path "env:\ProgramFiles(x86)") {
+          $pf32 = Get-Content "env:\ProgramFiles(x86)"
+      }
+
+      # Canonicalize so we can do StartsWith tests
+      if(!$pf.EndsWith("\")) { $pf += "\" }
+      if($pf32 -and !$pf32.EndsWith("\")) { $pf32 += "\" }
+
+      $UserHome = $RuntimeHomes.Split(";") | Where-Object {
```
dougbu: Added `Split()` and earlier `if ($RuntimeHomes)` because this block was currently after `$RuntimeHomes` was an array, not a string.

src/dnvm.ps1
```diff
@@ -196,40 +215,28 @@ if(!$GlobalHome) {
...
-    $env:DNX_HOME="$env:DNX_HOME;$env:DNX_GLOBAL_HOME"
-    $UnencodedHomes = "$UnencodedHomes;$GlobalHome"
+# Determine where runtimes can exist (RuntimeHomes)
+if(!$RuntimeHomes) {
+    # Set up a default value for the runtime home
+    $UnencodedHomes = "$UserHome;$GlobalHome"
+} elseif ($RuntimeHomes.StartsWith(';')) {
+    _WriteOut "Ignoring invalid $HomeEnvVar; value was '$RuntimeHomes'" -ForegroundColor $ColorScheme.Warning
+    Set-Content "env:\$HomeEnvVar" $null
+    [Environment]::SetEnvironmentVariable($HomeEnvVar, $null, "User")
```
dougbu: Note this doesn't clean much up when `dnvm.ps1` is invoked through `dnvm.cmd`. At least it doesn't _use_ the messy `$env:DNX_HOME` value anymore. Could make changes to the `$CmdPathFile` file within `Set-Path` but that's not always called.

src/dnvm.ps1
```diff
@@ -1543,17 +1550,17 @@ function dnvm-use {
 
     if ($versionOrAlias -eq "none") {
         _WriteOut "Removing all runtimes from process PATH"
-        Set-Path (Change-Path $env:Path "" ($RuntimeDirs))
+        Set-Path (Change-Path $env:Path "" $RuntimeDirs)
```
dougbu: Teensy nit I fixed -- `$RuntimeDirs` is always an array and the parens weren't needed.

src/dnvm.ps1
```diff
@@ -1705,13 +1712,13 @@ function dnvm-setup {
     if(Test-Path "env:\$HomeEnvVar") {
         $processHome = Get-Content "env:\$HomeEnvVar"
     }
-    $processHome = Change-Path $processHome "%USERPROFILE%\$DefaultUserDirectoryName" $PathsToRemove
+    $processHome = Change-Path $processHome "$DefaultUserHome" $PathsToRemove
```
dougbu: Not entirely sure how `%USERPROFILE%` gets expanded but it worked before and continues to work.

dougbu: Tested again. This sets the environment variable to `%USERPROFILE%\.dnx`. `dnvm.ps1` is fine with that now but it means `$env:DNX_HOME` doesn't contain a global home value.

@BrennanConroy should this instead set `$env:DNX_USER_HOME`? Since it's providing a default, might be better to leave the environment variables undefined...

src/dnvm.ps1
```diff
@@ -181,13 +181,32 @@ if($CmdPathFile) {
...
-if(!$RuntimeHomes) {
-    # Set up a default value for the runtime home
-    $UnencodedHomes = "$env:USERPROFILE\$DefaultUserDirectoryName;$GlobalHome\$DefaultGlobalDirectoryName"
-} else {
-    $UnencodedHomes = $RuntimeHomes
+# Determine the default installation directory (UserHome)
+if(!$UserHome) {
+    if ($RuntimeHomes) {
+      _WriteDebug "Detecting User Home..."
+      $pf = $env:ProgramFiles
```
BrennanConroy: We use AllUsersProfile and ProgramData below for global home, but here we use ProgramFiles and ProgramFiles(x86)

BrennanConroy: Oh, this is old code, maybe I don't know what it's doing?

dougbu: This code is looking thru `$env:DNX_HOME` for the first entry that is _not_ under `$env:\ProgramFiles(x86)` or `$env:ProgramFiles`. Not sure why it isn't also checking for `$env:ProgramData` but wasn't the default `$GlobalHome` previously `$env:ProgramFiles\Microsoft DNX`?

In any case, I didn't change the details of this code. Just when it runs.

BrennanConroy: Could we change this then?

src/dnvm.ps1
BrennanConroy: We shouldn't leave DNX_HOME without a global home value

src/dnvm.ps1
dougbu: @anurse your thoughts? You wrote this block back in 3266ead.

src/dnvm.ps1
dougbu: Agree the current code looks wrong. But why bother setting `$env:DNX_HOME` (or `$env:DNX_USER_HOME`) to its default? Script will figure same value whenever it's run.

In other words, let's use the environment variables for user _overrides_.

dougbu:
Updated; CI builds passing

src/dnvm.ps1
anurse: At the time it was written, the MSI was putting the `ProgramFiles` path into `DNX_HOME` so we wanted to avoid defaulting to that as the install location.

BrennanConroy:
:ship: :it:

dougbu:
416db7f

PR #451, opened 2015-09-23
adds support for the windows

slofurno:
this change allows dnvm.sh to run on windows
- adds a case to return "win" to __dnvm_current_os
- sets the DNX_USER_HOME environment variable to the correct folder on windows while also being compatible with the ":" delimiter used elsewhere

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -42,6 +42,10 @@ if __dnvm_has "unsetopt"; then
     unsetopt nomatch 2>/dev/null
 fi
 
+if [[ $(uname) == *"CYGWIN"* ]]; then
+    DNX_USER_HOME=$(cygpath "$HOMEPATH/$_DNVM_RUNTIME_FOLDER_NAME")
+fi
+
 if [ -z "$DNX_USER_HOME" ]; then
     eval DNX_USER_HOME="~/$_DNVM_RUNTIME_FOLDER_NAME"
 fi
@@ -67,13 +71,14 @@ DNX_ACTIVE_FEED=""
 __dnvm_current_os()
 {
     local uname=$(uname)
-    if [[ $uname == "Darwin" ]]; then
+    if [[ $uname == *"CYGWIN"* ]] || [[ $uname == *"MINGW"* ]]; then
+        echo "win"
+    elif [[ $uname == "Darwin" ]]; then
         echo "darwin"
     else
         echo "linux"
     fi
 }
-
 __dnvm_os_runtime_defaults()
 {
     local os=$1
@@ -929,7 +934,6 @@ dnvm()
             [[ ! -d $_DNVM_USER_PACKAGES ]] && echo "$_DNVM_RUNTIME_FRIENDLY_NAME is not installed." && return 1
 
             local searchGlob="$_DNVM_RUNTIME_PACKAGE_NAME-*"
-
             local runtimes=""
             for location in `echo $DNX_HOME | tr ":" "\n"`; do
                 location+="/runtimes"


src/dnvm.sh
```diff
@@ -46,6 +57,10 @@ if [ -z "$DNX_USER_HOME" ]; then
     eval DNX_USER_HOME="~/$_DNVM_RUNTIME_FOLDER_NAME"
 fi
 
+if [ $(__dnvm_current_os)=="win" ]; then
+    eval DNX_USER_HOME="/c/Users/$USERNAME/$_DNVM_RUNTIME_FOLDER_NAME"
```
pranavkm: Shouldn't this be using `%UserProfile%`?

src/dnvm.sh
```diff
@@ -848,10 +855,11 @@ dnvm()
             [[ ! -d $_DNVM_USER_PACKAGES ]] && echo "$_DNVM_RUNTIME_FRIENDLY_NAME is not installed." && return 1
 
             local searchGlob="$_DNVM_RUNTIME_PACKAGE_NAME-*"
-
             local runtimes=""
             for location in `echo $DNX_HOME | tr ":" "\n"`; do
                 location+="/runtimes"
+                
+                echo $location
```
pranavkm: Undo.

src/dnvm.sh
```diff
@@ -2,8 +2,8 @@
 # Source this file from your .bash-profile or script to use
 
 # "Constants"
-_DNVM_BUILDNUMBER="{{BUILD_VERSION}}"
-_DNVM_AUTHORS="{{AUTHORS}}"
+_DNVM_BUILDNUMBER="beta8-15516"
```
pranavkm: Undo both these lines.

src/dnvm.sh
```diff
@@ -30,6 +30,17 @@ if [ "$NO_COLOR" != "1" ]; then
     Whi='\e[0;37m';     BWhi='\e[1;37m';    UWhi='\e[4;37m';    IWhi='\e[0;97m';    BIWhi='\e[1;97m';   On_Whi='\e[47m';    On_IWhi='\e[0;107m';
 fi
 
+__dnvm_current_os()
+{
+    local uname=$(uname)
+	if [ "$(expr substr $(uname -s) 1 10)" == "MINGW32_NT" ] || [ "$(expr substr $(uname -s) 1 9)" == "CYGWIN_NT" ]; then
```
pranavkm: Fix indentation.

pranavkm:
Could you also remove all the extra new lines that were added?

src/dnvm.sh
slofurno: $USERPROFILE has the value C:\Users\<username>, while this produces /c/Users/<username>.

pranavkm: You can't rely on it being `C:\users` though. Is there another variable that could be used?

slofurno: you could convert userprofile to the unix style directory

replaced=${USERPROFILE//\\/\/}
eval DNX_USER_HOME="/${replaced/:\//\/}/$_DNVM_RUNTIME_FOLDER_NAME"

seems to give the same output

pranavkm: Feels too hacky. I don't feel comfortable putting that code here. Can we instead use `$HOME` when you're in Cygwin? It defaults to the user profile directory by default. cc @davidfowl

davidfowl:
I still want to understand why we need this...

pranavkm:
@anurse thoughts on what it would take to get the sh tests to run on Cygwin \ MinGW?

anurse:
We'd have to install Cygwin/MinGW on the CI machines... but otherwise it should work.

As to @davidfowl 's comment, I can see the use case (using `dnvm` from inside Cygwin/MinGW) but I do think it's a pretty small use case. I'm not sure if we want to take on the work of setting up the test infrastructure. If the changes are minor, I suppose we could take them and unofficially support it (i.e. we'll support it if we can but no guarantees and don't call support ;)).

slofurno:
im sorry if i overestimated the usefulness of this change, I expected that many of the same c# developers that are currently using the coreclr to run asp.net 5 applications on both windows and linux would be using cygwin since it offers a consistent environment for cross platform development.

anurse:
I'm sure there are developers out there doing it, we just can't really commit resources to testing it and fully supporting it :). I'm certainly OK with making some small changes to support it and these look relatively benign

src/dnvm.sh
anurse: Aside from the indentation, I think wildcards are the more idiomatic way to do this in sh. Something like `if [[ "$(uname -s)" == *"MINGW32_NT"* ]] ...`. As long as that still works in Cygwin I'd suggest switching to that.

src/dnvm.sh
anurse: This is going to get even more complicated if we take https://github.com/aspnet/dnx/issues/1507

muratg:
@davidfowl One person reported their need a while back: https://github.com/aspnet/dnvm/issues/280

muratg:
@davidfowl Trouble is for cygwin or msys bash users, even dnvm.cmd doesn't work. :/

davidfowl:
@muratg as long as we take on the testing burden and make it match the quality of the other scenarios I'm all for it.

muratg:
@davidfowl agreed, and I'm not saying we should do it. Was just answering your question.

muratg:
@davidfowl your experimental branch can be the real fix for this ;)

davidfowl:
@muratg I'll finish it one day.

anurse:
@davidfowl @muratg I think there is still value to small changes like this even if we decide not to take on the burden of testing Cygwin/MSys support. We can let that be community-supported. As long as the changes don't break standard Windows/Linux scenarios and are relatively minor, I say let 'em happen.

pranavkm:
:+1:

slofurno:
i setup a ci with the following config https://github.com/slofurno/dnvm-ci/blob/master/appveyor.yml

pranavkm:
@anurse \ @davidfowl what're we doing with this given the discussion https://github.com/aspnet/dnx/issues/3068#issuecomment-151560270?

muratg:
We can discuss post RC1.

anurse:
If it doesn't break our existing scenarios, I don't have a problem taking these and making dnvm.sh-on-cygwin a "community maintained" feature

jasonwilliams200OK:
:+1:  for this.
Would be really cool if we put a table somewhere in docs to indicate which shells are supported (with "community maintained" represented by `*` with a legend note). This will help the discoverability and people will keep testing with all those scenarios and provide feedback if some feature horked on some rarely-used shell environment. We can also hookup couple of tests with CI as AppVeyor provides Cygwin and MinGW, in which case if something breaks and the resolution is not obvious (which is very unlikely), aspnet team can ping interested community members (such as me) to suggest the resolution.

muratg:
As discussed in the [Roadmap](https://github.com/aspnet/Home/wiki/Roadmap) we will be moving to [Dotnet CLI](https://github.com/dotnet/cli) in ASP.NET 5 RC 2 and retire DNX, DNU and DNVM command line tools.  Closing this PR.

Thank you for your contribution!

atrauzzi:
@muratg - Will the tickets related to this functionality be transitioned there?

muratg:
@atrauzzi Yes the relevant tickets are moved to dotnet CLI, CoreFX, CoreCLR or NuGet repos as needed. About this particular one, I think the team is currently thinking of not using .sh & .ps1 but actually writing the dnvm equivalent in managed code. Thanks for the bug report on the CLI to keep this in the radar too!

cataquil:
I think slofurno has a point, those dotnet devs who have gone polyglot
with oss crosspolination, nodejs, and also and thriving npm popularity,
and the retro coolness of using the command line.
found it was either git-bash, buy a mac, or run in a linux vm or docker
container.

Many OSS projects supported posix/unix command lines and so to learn 
from those communities git-bash/msys was the only way.

slofurno script works, however dnu isn't found, this happens on CMD,
running thru the cmder terminal. of course its found in default:Start->run

dnvm upgrade -a x64 -r coreclr
usually wired up dnu.

PR #452, opened 2015-09-24
Added 'dnvm alias -d' and deprecated 'dnvm unalias'

### LINKED ISSUE #418: dnvm alias -d (not working in Linux) also missing help
Hi there,

The following command

dnvm alias -d default

works in Windows, with the following output

Removing alias default

In Linux, the very same command gives the default framework an alias of 'd'.

Also the command dnvm alias help doesn't produce any help, instead it says

"There is no alias called 'help'"

mholo65:
Fixes #418 

//cc @BrennanConroy @anurse


dougbu:
:shipit: I'll get this in shortly

dougbu:
16530be

PR #467, opened 2015-10-03
Aligned location in dnvm list -detailed on Linux

### LINKED ISSUE #458: dnvm list -detailed does not align locations on linux
On Windows `dnvm list -detailed` produces this:
image

Note that the locations are aligned.

On Linux the same command produces this:

image

Note that the location is no longer aligned for default.

mholo65:
Fixes #458 

//cc @BrennanConroy @muratg


BrennanConroy:
Looks good, I'll merge this soon

PR #469, opened 2015-10-05
-r, -a and -OS parameters are now case insensitive

### LINKED ISSUE #449: Alias -r is case sensitive
`dnvm alias default -r CoreCLR` will create an alias but then not work correctly due to the case.

It should either fail and tell me that the casing is wrong or actually work. Ideally it would work, perhaps lowering all the runtime names since they are all lower case in the file names.

mholo65:
```__dnvm_requested_version_or_alias``` parameters are now case insensitive and dnvm alias can now take 9 parameters.

This fixes #449 and also other issues caused by case sensitivity in commands using ```__dnvm_requested_version_or_alias```

Was this also an issue on Windows? Not currently near a Windows machine so I can't verify.

//cc @muratg @glennc @BrennanConroy

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -38,6 +38,10 @@ __dnvm_has() {
     return $?
 }
 
+__dnvm_to_lower() {
+    echo "$1" | tr '[:upper:]' '[:lower:]'
+}
+
 if __dnvm_has "unsetopt"; then
     unsetopt nomatch 2>/dev/null
 fi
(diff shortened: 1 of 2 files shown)


mholo65:
Should also functions such as ```__dnvm_os_runtime_defaults```, ```__dnvm_runtime_bitness_defaults```, ```__dnvm_find_latest``` and ```__dnvm_find_package``` have case insensitive parameters? Now commands like e.g. ```dnvm install latest -r CoreCLR``` and ```dnvm install 1.0.0-beta7 -r CoreCLR``` will fail.

muratg:
Yeah, to me it looks like install/upgrade should behave the same way when it comes to -r

cc: @glennc

mholo65:
I suppose it should be same behavior when it comes to parameters ```-a|-arch``` and ```-OS```. At least Windows (ps1) treats all these parameters as case insensitive.

muratg:
Cool, thanks. Could you send an update? :smile:

mholo65:
Refactored, and rebased. 

Added function ```__dnvm_to_lower``` which is called on all occurrences of ```-r|-runtime```, ```-a|-arch``` and ```-OS``` parameters.

This should now be consistent with Windows.

muratg:
Awesome, thanks!

muratg:
@BrennanConroy please review and merge at your convenience. :)

glennc:
Really happy that we got a PR with a test included :+1:

muratg:
Agreed with @glennc , thanks @mholo65!   @BrennanConroy will merge this in when he's back in the office.

Issue #439, opened 2015-09-17
dnvm setup should set DNX_HOME machine-wide on Windows


anurse:
When we install DNVM on Windows, we should set the `DNX_HOME` environment variable to the appropriate value (including global install dirs). This should be in the `dnvm-setup` function, and should run only if elevated.

The MSI should (already?) set the same environment variable so this would only be for scenarios where DNVM is installed via the "copy-pasta" script.

/cc @marcind @BrennanConroy @glennc

marcind:
Make sure the MSI is available via Web Platform Installer.

anurse:
@danroth27 @DamianEdwards Are there plans to integrate our MSIs into WebPI?

glennc:
I haven't talked about adding it to WebPI with anyone yet, but it should be possible. I take it you like WebPI and want to use that to install all of our stuff?

@anurse this is just to prep the machine for wherever we may drop DNXs in the future?

marcind:
@glennc I'm looking for any method that allows me to install dependencies onto a target host in a 'one-liner' declarative, idempotent fashion (i.e. I can automate this). WebPI and Chocolatey fit that bill as I can just choose the product and version and they take care of the rest (including no-op if it's already installed). No need to look up download links, manage MSIs, etc. So if you're going to have an MSI, make sure to add it to those registries.

Though honestly, I've been quite happy with the capabilities of DNVM so far (esp. once the machine-wide install story gets completed).

anurse:
We've also had people ask about PowerShell Package Management/Desired State Configuration support which is compatible with MSIs.

In general, I think a major RC-level thing will be refining our acquisition story so feedback here is welcome :). In this case, `dnvm` could have (and basically does already) it's own idempotent "machine setup" script so that all you have to do is bring `dnvm.ps1` to the box, run `setup` and maybe `update-self` to get the true latest (if you didn't directly download the latest) and you should be set.

marcind:
Since we're talking about covering all the bases here, having this available via `apt-get` on Ubuntu would also be great :)

DamianEdwards:
@marcind we're working on that right now. We plan to have the following acquisition experiences ready for RC in November (some may come online earlier):
- Scripted install for all platforms (curl a script and run it to get DNVM)
- DNVM for all platforms (including support for global/shared installs)
- EXE installer for Windows (with CLI options to specify what gets installed)
- PKG installer for OSX
- apt-get feed for Ubuntu and related distros
[...]

PR #473, opened 2015-10-09
Set machine-wide DNX_HOME in dnvm-setup.

cesarbs:
Set machine-wide DNX_HOME environment variable when running dnvm setup on Windows in elevated mode.

#439

### DIFF
--- a/src/dnvm.ps1
+++ b/src/dnvm.ps1
@@ -1762,7 +1762,8 @@ function dnvm-exec {
 #>
 function dnvm-setup {
     param(
-        [switch]$SkipUserEnvironmentInstall)
+        [switch]$SkipUserEnvironmentInstall,
+        [switch]$SkipSetMachineDnxHome)
 
     $DestinationHome = [Environment]::ExpandEnvironmentVariables("$DefaultUserHome")
 
@@ -1794,6 +1795,13 @@ function dnvm-setup {
         $userPath = [Environment]::GetEnvironmentVariable("PATH", "User")
         $userPath = Change-Path $userPath $Destination $PathsToRemove
         [Environment]::SetEnvironmentVariable("PATH", $userPath, "User")
+    }
+
+    if(Is-Elevated) {
+        if (!$SkipSetMachineDnxHome) {
+            _WriteOut "Adding Machine $HomeEnvVar"
+            [Environment]::SetEnvironmentVariable($HomeEnvVar, $Destination, [System.EnvironmentVariableTarget]::Machine)
+        }
     }
 
     # Now clean up the HomeEnvVar if currently set; script installed to default location.


cesarbs:
@anurse @glennc  @muratg

cesarbs:
I tried adding unit tests, but I couldn't find a clean way to mock Is-Elevated and setting a machine environment variable.

src/dnvm.ps1
```diff
@@ -1796,6 +1797,13 @@ function dnvm-setup {
         [Environment]::SetEnvironmentVariable("PATH", $userPath, "User")
     }
 
+    if(Is-Elevated) {
+        if (!$SkipSetMachineDnxHome) {
```
dougbu: Combine this condition into the previous `if`.

src/dnvm.ps1
```diff
@@ -1796,6 +1797,13 @@ function dnvm-setup {
...
+            _WriteOut "Adding Machine $HomeEnvVar"
+            [Environment]::SetEnvironmentVariable($HomeEnvVar, $Destination, [System.EnvironmentVariableTarget]::Machine)
```
dougbu: This function installs `dnvm.cmd` in `$env:DNX_USER_HOME` which is only part (half) of the default `$env:DNX_HOME`. Should set `$env:DNX_USER_HOME` here.

src/dnvm.ps1
dougbu: Also the default should be _not_ updating the machine-wide environment variables. It may be useful but will cause problems when multiple users run `dnvm` on the same machine. Per-user `dnvm setup`s do not set `$env:DNX_USER_HOME` (or `$env:DNX_HOME`). (We decided it was not normally useful to define environment variables for defaults this script knows.) So every user will pick up that one user's installation location which is probably not what they want.

anurse: Actually I think we do want this to be the default, but we also need a small change here. DNX_HOME itself supports using `%ENV%` tokens to represent environment variables. The intent here is for `DNX_HOME` to have a value of `%USERPROFILE%\.dnx;[Machine Global install location]` so that it will expand when each user uses it. I would also suggest that if `DNX_HOME` is already set, we not modify it.

The purpose here is to ensure someone sets this because without it we have various scripts that search `DNX_HOME` which will not work in the default case.

src/dnvm.ps1
anurse: Why? This is only about setting `DNX_HOME` which is the search paths for other scripts to use to locate DNX. `DNX_USER_HOME` is only for overriding the default user-level installation location, so a user only needs to set it if they are changing it.

src/dnvm.ps1
dougbu: A machine-wide value of `%USERPROFILE%\.dnx;[Machine Global install location]` set using `SetEnvironmentVariable()` might not work work if that creates a `REG_EXPAND_SZ` entry. The system expands such environment variables as it starts up. Then `%USERPROFILE%` would work correctly only for per-user environment variables.

Need to test the resulting key type.

src/dnvm.ps1
dougbu: @anurse what other scripts? That said, might be :ok: if this code set `DNX_HOME` _correctly_ and avoided the `REG_EXPAND_SZ` issues I mentioned above.

src/dnvm.ps1
anurse: Ooh, right, I remember seeing that before. Hrm...

anurse:
@dougbu @cesarbs and I chatted and we think there are some bigger issues. I'm going to summarize that discussion over on the issue (#439) but we can probably close this PR for now.

cesarbs:
Closing PR until we get a better design on this.

PR #476, opened 2015-10-16
Push DNVM to blob storage

shirhatti:
Upload to blob storage by setting DNVM_DEPLOY_CONTAINER to the desired Azure Blob storage container with Shared Access Signature

### DIFF
--- a/build/_push-dnvm-blob.shade
+++ b/build/_push-dnvm-blob.shade
@@ -1,0 +1,31 @@
+use assembly="System.Net.Http, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"
+
+use namespace='System'
+use namespace='System.IO'
+use namespace='System.Net.Http'
+
+@{
+    var queryString = new Uri(container).Query;
+    var baseAddress = container.Substring(0, container.IndexOf('?')) + "/";
+    
+    var client = new HttpClient()
+    {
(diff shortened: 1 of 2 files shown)


pranavkm:
:shipit:

Eilon:
What exactly is this for? What blob feed is this? Who consumes it?

shirhatti:
CC @glennc 
The idea is to push `dnvm` to a storage account sitting behind a CDN for the new acquisition experience (get.asp.net)

Eilon:
When do we intend this to run? After every build of DNVM, or after Coherence-Signed passes?

pranavkm:
@Eilon this would run after every successful build of DNVM. We do the same thing when publishing dnvm to the Home repo.

Eilon:
@pranavkm but these builds aren't tested. They're just the latest build no matter what. Who will consume these?

glennc:
This is the same as it is today for the home repo. We push to the home repo after every build and if you run dnvm upgrade-self you get the latest dnvm.

Putting them on blob storage does give us the potential to change this and have dnvm be in channels the same a the runtimes/installers are. But right now this is an exact copy of what we have right now, but using blob storage instead of github as a vehicle.

glennc:
I am fine with changing this to push to /dev/dnvm.sh so that we can make channels first class for dnvm as well.

Eilon:
Hmm I guess I'm just wondering what the long term plan is for this.

PR #478, opened 2015-10-20
dnvm returns non zero if no argument is passed

bjorkstromm:
As per the discussion in aspnet/dnx#2920 @davidfowl said that `dnx` and `dnu` returns non zero exit code if no argument is provided because:
1. other well known programs do it
2. We had a bug once where the arguments changed and help was being output on the CI but nothing was failing

To be consistent, I think that also `dnvm` should have the same behaviour? I've only done Linux, but if You like the idea, I'm willing to also fix Windows.

//cc @moozzyk @BrennanConroy

### DIFF
--- a/src/dnvm.sh
+++ b/src/dnvm.sh
@@ -504,7 +504,7 @@ dnvm()
 
         printf "%b\n" "Use ${Yel}$_DNVM_COMMAND_NAME [help|-h|-help|--help] ${RCol} to display help text."
         echo ""
-        return
+        return 2
     fi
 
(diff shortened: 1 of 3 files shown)


muratg:
As discussed in the [Roadmap](https://github.com/aspnet/Home/wiki/Roadmap) we will be moving to [Dotnet CLI](https://github.com/dotnet/cli) in ASP.NET 5 RC 2 and retire DNX, DNU and DNVM command line tools.  Closing this PR.

Thank you for your contribution!

PR #481, opened 2015-10-26
Fix concurrent dnvm

BrennanConroy:
@pranavkm @pakrym @dougbu

### DIFF
--- a/test/apps/TestApp/project.json
+++ b/test/apps/TestApp/project.json
@@ -10,11 +10,11 @@
         "dnx451": { },
         "dnxcore50": {
             "dependencies": {
-                "System.Runtime": "4.0.0-*",
-                "System.Console": "4.0.0-*",
-                "System.Collections": "4.0.10-*",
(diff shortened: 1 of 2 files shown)


pakrym:
:shipit:

test/apps/TestApp/project.json
```diff
@@ -11,7 +11,7 @@
         "dnxcore50": {
             "dependencies": {
                 "System.Runtime": "4.0.0-*",
-                "System.Console": "4.0.0-*",
+                "System.Console": "4.0.0-beta-23225",
```
dougbu: Huh?

BrennanConroy: I think we should pin all the versions since this is a test app for a specific version of dnx, with 4.0.0-* it picked up a package with runtime.json and the beta7 dnx doesn't understand those.

dougbu: Ah :ok_hand:, just seemed strange in this PR.

dougbu:
:shipit:

PR #487, opened 2015-11-10
Create user package directory on DNVM source

anurse:
This fixes a bug where DNVM won't show globally installed DNXes if the user runtimes path is not present

/cc @glennc @BrennanConroy


muratg:
Looks good. :shipit: once we have the approval.

Issue #190, opened 2015-03-25
[blocking]: Unable to install dnvm (optimistic) when user name has a space


sayedihashimi:
From aspnet/home I pasted the install script into a cmd window on Windows where my username is "Sayed Hashimi" and ran into the following error.

```
C:\temp\_NET\WebApplication3\src\WebApplication3>@powershell -NoProfile -ExecutionPolicy unrestricted -Command "&{$Branch='dev';iex ((new-object net.webclient).DownloadString('ht
tps://raw.githubusercontent.com/aspnet/Home/dev/dnvminstall.ps1'))}"
Using temporary directory: C:\Users\SAYEDH~1\AppData\Local\Temp\dnvminstall
Downloading DNVM.ps1 to
Downloading DNVM.cmd to
Installing DNVM
WARNING: Found a KRE_HOME environment variable. This variable has been deprecated and should be removed, or it may interfere with DNVM and the .NET Execution environment
Installing .NET Version Manager to C:\Users\Sayed Hashimi\.dnx\bin
Installing 'dnvm.ps1' to 'C:\Users\Sayed Hashimi\.dnx\bin' ...
Installing 'dnvm.cmd' to 'C:\Users\Sayed Hashimi\.dnx\bin' ...
Adding C:\Users\Sayed Hashimi\.dnx\bin to Process PATH
Adding C:\Users\Sayed Hashimi\.dnx\bin to User PATH
Adding C:\Users\Sayed Hashimi\.dnx to Process DNX_HOME
Adding C:\Users\Sayed Hashimi\.dnx to User DNX_HOME
'""C:\Users\Sayed' is not recognized as an internal or external command,
operable program or batch file.
The system cannot find the path specified.
```

Working around this is tricky because the scripts download other scripts which need to be modified. This is a really bad experience and I think usernames with a space may be common (don't have any data on that).

wangkanai:
Having the same problem here too #192

sayedihashimi:
@damianedwards @davidfowl this is a blocking issue for users with a space in their username. Can someone take a look? I think it's a regression as week.

muratg:
@glennc looks like more folks are hitting this. Could you take a look when you have time? Otherwise, I'll ask @anurse :)

sayedihashimi:
If we can't get it fixed soon we should at least have a documented workaround.

janna173:
I am getting the same error although this seemed to be fixed?!
Any suggestions what to do?

[...]

PR #493, opened 2015-11-18
Fix space in username issue

BrennanConroy:
Should fix the problems at the end of this issue #190 

@anurse

### DIFF
--- a/src/dnvm.cmd
+++ b/src/dnvm.cmd
@@ -1,6 +1,6 @@
 @Echo off
 
-for /f "delims=" %%i in ('PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.IO.Path]::GetTempFileName()"') do set DNVM_CMD_PATH_FILE=%%i.cmd
+for /f "delims=" %%i in ('PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.IO.Path]::GetTempFileName()"') do set DNVM_CMD_PATH_FILE="%%i.cmd"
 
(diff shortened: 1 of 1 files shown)


anurse:
:shipit:

danielcrenna:
This is still happening with the published RC1 bits. Do you know when this will make it into the distro?

BrennanConroy:
This is fixed in RC2 bits, `dnvm update-self`

danielcrenna:
The issue with that is that the broken spaces actually prevents `dnvm update-self` for the same reason.

Where the user is "Daniel Crenna", `update-self` fails

BrennanConroy:
You'll have to manually update then.

danielcrenna:
Sure, where are the bits?

BrennanConroy:
https://github.com/aspnet/Home/dnvm.cmd and https://github.com/aspnet/Home/dnvm.ps1 but for this issue specifically you only really need to replace the cmd

danielcrenna:
Yes, that was the ticket. Thank you for leaving that there for myself and
others.

rclabo:
I have the same problem due to a space in my username.  I need to manually update my dnvm.cmd file apparently.  the links that @BrennanConroy posted a couple weeks ago now return 404 errors.  Where can I get the dnvm.cmd file now to manually update mine?

danielcrenna:
@rclabo they always 404'd but you can hit the main repository that the link
alludes to and find them in root directory

rclabo:
@danielcrenna Thanks!

danielcrenna:
@rclabo Looks like they moved things around slightly but might have luck at https://github.com/aspnet/dnvm/tree/dev/src

PR #514, opened 2016-03-26
Using find -o instead of -or

hultqvist:
BusyBox `find` didn't recognize -or but -o

According to the find manual -or is not POSIX compliant.

Before this change commands such as `dnvm list` failed.


muratg:
Hi @hultqvist, thanks for the PR!  We're retiring dnvm/dnx/dnu in RC2 so we're not making any more changes in this area. Closing this one.

